PR #9, opened 2011-09-28
Support attaching notes when tracking events

esad:
Adds new track:note: and track:properties:note: methods, which internally attach a "mp_note" property to the event.


samgreen:
This is now implemented in master.

PR #12, opened 2011-11-11
A few changes to make the library work in desktop apps + a sample app.

fpotter:
There's not much to this.  The biggest change is that on Mac, the archived events have to be stored under ~/Library/Application Support/AppName.

Verified that the archive/unarchive of the event queue works just as it does on iOS.  Verified that events are firing in both iOS and Mac.

**Sample App Screenshot:**


samgreen:
The best approach for using Mixpanel in an OSX app is to use the WatchExtension subspec of the CocoaPod.

Closing this PR.

PR #15, opened 2012-01-05
Added an additional function removeSuperPropertyForKey and also an assert for checking when the postBody become empty

deepakprabhakara:
- Added removeSuperPropertyForKey
- Added NSAssert to ensure postBody is not empty due to failed encoding


samgreen:
Closing this stale PR.

PR #66, opened 2013-07-09
Added stopFlushTimer to the header

eperezrico:
You must call stopFlushTimer before replacing a retained Mixpanel object, otherwise the timer will retain the pointer, 
 causing dealloc not to be called and therefore the object not being removed as observer, causing the notification center to access released memory.


samgreen:
Closing this stale PR.

PR #68, opened 2013-07-17
Support for MacOS X Desktop application

lcerveau:
As I need MixPanel capabilities in a Desktop application, I did modify the iOS code to be able to compile and execute under MacOSX. The first test are positive but review is certainly needed to explore all MixPanel features. Following points have to be noted
- distinctID is the serial number of the desktop machine and is taken using IOKit method
- depending if the application is sandboxed or not the archive location may differ
- All UIApplication related notifications have been "transferred" to their NSApplication couterparts
- Currently the code considers background handling, but it may be more suited to not stop the flush while in the background. This part may need more work
This is a first version of this support. As code changes start to be "big" , I am asking for a pull request and ad a review process before going further, in order to get feedback as soon as possible in the process.
thanks


samgreen:
See #12

PR #100, opened 2014-01-22
Set a locale and calendar on NSDateFormatter

peymano:
Avoid device-specific locale and calendar settings.

See https://developer.apple.com/library/ios/qa/qa1480/_index.html


samgreen:
This is currently resolved in the master branch.

PR #124, opened 2014-04-14
Call startFlushTimer in initWithToken:andFlushInterval:. Fixes #117

### LINKED ISSUE #117: flushTimer is not started after init
Invocation of ```startFlushTimer``` in ```initWithToken:andFlushInterval:``` was removed in this commit e5d85f14d695be2fb9a3e28dc59fc9197a659ec9.
If the library is initialized not in application:didFinishLaunchingWithOptions:, flushing is delayed until the application enters background. This happens if some wrapper like https://github.com/segmentio/analytics-ios is used. Segment.io stores API keys on the server. When application:didFinishLaunchingWithOptions: is called our application cannot init Mixpanel, because we have to retrieve the Mixpanel API key from the Segment.io server and this request takes some time.
Starting flushTimer at the end of ```initWithToken:andFlushInterval:``` could be a possible fix.

realf:
Hi guys. 
This is a possible fix for #117. It would be great if Mixpanel would not rely on initializing in applicaiton:didFinishLaunchingWithOptions: at all, but this fix is pretty good too.
Cheers


samgreen:
Closing this stale PR

PR #143, opened 2014-07-10
Adds Prefix Header and Xcode Project

### LINKED ISSUE #142: Support for importing Mix Panel as a subproject/dependency
Thanks for making the SDK available via Git! That makes it really easy to bring in as a submodule, or flow down into our source control if we want to try to make patches. But there are a few problems...

- The import process is kind of awful if one isn't using CocoaPods and is using a pure Git submodule setup. I have to import all the source directly into my project, including a second asset catalog. Can a project have two asset catalogs? I'm not entirely sure, I guess I'll find out.
- Without the second project, the UI to control which Git remote/branch you're using for a dependency in Xcode doesn't work at all. With the normal dependency setup where I reference another project as a dependency, I can toggle where the branch is being pulled from for that dependency. Without a sub project, Xcode can't track that.
- This also seems like a nice setup for frameworks in future versions of iOS. The best way to handle frameworks is to have an Xcode project I can reference from my project, which does the building and then imports the framework into my bundle. This also solves the duplicate assets problem by bundling the assets into your framework.

It looks like I could use CocoaPods to generate a workspace, but that just seems like a mess considering that none of our build systems have CocoaPods installed, and if I did a one-off, I'd have to do it every time I reconfigured the branches. Blech.

What I'd like to see is really just an xcodeproj present in MixPanel that I can do a one shot import of into my xcodeproj.

jleishman:
- Prefix header includes Foundation and UIKit frameworks.
- Xcode project defines one target to build static library.
- Adds headers to Copy Files build phase.

Please check whether I'm missing any frameworks or libraries that need to be linked with the targets. I think the prefix header is a good thing, but it's also kind of duplicating some of the work done by Mixpanel.h. I would be inclined to pull that out into the prefix header, but that may be inconvenient for other users.

Fixes #142

### DIFF
--- a/Mixpanel.xcodeproj/project.pbxproj
+++ b/Mixpanel.xcodeproj/project.pbxproj
@@ -1,0 +1,387 @@
+// !$*UTF8*$!
+{
+	archiveVersion = 1;
+	classes = {
+	};
+	objectVersion = 46;
+	objects = {
+
+/* Begin PBXBuildFile section */
+		D0796154196E95C300BC626E /* Foundation.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = D0796153196E95C300BC626E /* Foundation.framework */; };
+		D0796194196E962A00BC626E /* MPNotification.m in Sources */ = {isa = PBXBuildFile; fileRef = D079617D196E962A00BC626E /* MPNotification.m */; };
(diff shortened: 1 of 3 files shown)


samgreen:
We have opted to remove PCH files in favor of CocoaPods.

PR #147, opened 2014-07-23
Allow developer to disable location tracking

sadlerjw:
MixPanel collects the user's approximate location data based on IP
address; this change allows the developer to opt out of this data
collection.


samgreen:
This PRQ is not enough to prevent this behavior. Additionally any user of the library can prevent this behavior by passing a custom event property like `{ @"ip": @NO }`.

Closing this issue.

PR #157, opened 2014-09-02
Adds a reset method with an optional completionBlock

hwaxxer:
When a user signs out of the app, we call `-reset`. However, if we immediately call `-identify:` with a new id that value is subsequently overwritten since `-reset` dispatches its work asynchronously.

Usage:

```
[[Mixpanel sharedInstance] resetWithCompletion:^{
    [[Mixpanel sharedInstance] identify:@"some id"];
}];
```


samgreen:
Closing this stale PR.

PR #159, opened 2014-09-02
Fixes #158. Optionally allows a user to pass in an app window

### LINKED ISSUE #158: Assumes window lives on the app delegate
On this line, the Mixpanel SDK makes assumptions about how windows are presented in the app. This is a glaring violation of private data. Let us recall that the window is private state, even if the Xcode templates expose it as public. This means, you shouldn't depend on its location, as it may not even be on the app delegate. For instance, in all of my apps, the main UIWindow (let us forget about any additional windows that may be in use, like on secondary displays) lives on a separate controller object which manages application state. The app delegate merely forwards along delegate calls, as is intended.

The offending line is here: https://github.com/mixpanel/mixpanel-iphone/blob/780c9dfe46ac69d83af8cf015d1f75e103de97bd/Mixpanel/Mixpanel.m#L287

My suggestion to correct this is require additional setup to hint Mixpanel at which window to use for this purpose. Let me give you a reference which you can hold onto weakly.

jeremytregunna:
Do not assume the window to attach the gesture recognizer to lives on the application delegate. Instead, allow the user to optionally pass in a reference to a window at creation time, hold a weak reference to that, and only fall back to the application delegate if that reference is nil.


samgreen:
Closing this stale PR.

PR #190, opened 2014-10-29
Implement Mixpanel in iOS project that uses Swift.

yiqin:
Cocoapods is not ready for Swift yet. When I implemented Mixpanel into my iOS project that uses Swift, I got several errors.
screen shot 2014-10-29 at 2 55 30 pm

The bugs come from #import UIKit/UIKit.h. The reason is in Objective-C project, we just need to add 
```Objective-C
#ifdef __OBJC__
	#import <UIKit/UIKit.h>
	#import <Foundation/Foundation.h>
#endif
```
in Prefix.pch

However, we don't do this in Swift project and we don't need to add framework one by one. So we have to add import UIKit/UIKit.h in necessary files. 

Here is one of file we need to import UIKit/UIKit.h
screen shot 2014-10-29 at 2 56 06 pm

We also need to add libicucore.dylib, CFNetwork.framework, and Security.framework in Link Binary With Libraries.


barrettclark:
Add the libraries as noted in the podfile, and also that you noted (libicucore.dylib, CFNetwork.framework, and Security.framework).  Then create a bridging header file (https://bohemianpolymorph.wordpress.com/2014/07/11/manually-adding-a-swift-bridging-header/) and put the `#import "Mixpanel.h"` in there.

Establish the shared instance in `didFinishLaunchingWithOptions` in `AppDelegate.swift`:
`Mixpanel.sharedInstanceWithToken("secretgibberish")`

From there you can track things in your Swift code:
```
let mixpanel = Mixpanel.sharedInstance()
mixpanel.track("Some View", properties: ["Some ID": some_id])
```

samgreen:
Closing this stale PR.

PR #209, opened 2015-01-09
Fix imports to be able to build static library and generate Xamarin bindings

NinZine:
Hi Mixpanel team!

Thanks for a clean and simple API! I have created bindings for Xamarin.iOS, in another repository. There were some problems building a static library from this code base, so attached are those changes that I had to make for it to work. Would of course like them to be included so the Xamarin bindings can easily follow updates from upstream.

Regards,
Andreas


malectro:
HI @NinZine , we got something very similar out in our latest release v2.7.1 . Thanks a ton for making it easy.

NinZine:
Confirmed working. Updated https://github.com/NinZine/mixpanel-xamarin

Thanks!

PR #217, opened 2015-01-21
Simulator autoconnect

malectro:
* automatically connects to the simulator when in debug mode
* also adds a flag to prevent connections

### DIFF
--- a/Mixpanel/MPABTestDesignerConnection.m
+++ b/Mixpanel/MPABTestDesignerConnection.m
@@ -34,6 +34,7 @@ NSString * const kSessionVariantKey = @"session_variant";
      */
     BOOL _open;
     BOOL _connected;
+    BOOL _keepTrying;
     NSURL *_url;
     NSMutableDictionary *_session;
     NSDictionary *_typeToMessageClassMap;
@@ -44,7 +45,7 @@ NSString * const kSessionVariantKey = @"session_variant";
     void (^_disconnectCallback)();
 }
 
-- (id)initWithURL:(NSURL *)url connectCallback:(void (^)())connectCallback disconnectCallback:(void (^)())disconnectCallback
+- (id)initWithURL:(NSURL *)url keepTrying:(BOOL)keepTrying connectCallback:(void (^)())connectCallback disconnectCallback:(void (^)())disconnectCallback
 {
     self = [super init];
     if (self) {
@@ -60,6 +61,7 @@ NSString * const kSessionVariantKey = @"session_variant";
 
         _open = NO;
         _connected = NO;
+        _keepTrying = keepTrying;
         _sessionEnded = NO;
         _session = [[NSMutableDictionary alloc] init];
         _url = url;
@@ -78,7 +80,7 @@ NSString * const kSessionVariantKey = @"session_variant";
 
 - (id)initWithURL:(NSURL *)url
 {
-    return [self initWithURL:url connectCallback:nil disconnectCallback:nil];
+    return [self initWithURL:url keepTrying:NO connectCallback:nil disconnectCallback:nil];
 }
 
 - (void)open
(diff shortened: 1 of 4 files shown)


Mixpanel/Mixpanel.h
```diff
@@ -201,6 +201,18 @@
...
  @property
+ 
+ @abstract
+ Controls whether to allow the lib to make connections to the Mixpanel
+ socket server regardless of the state of the DEBUG flag.
+ 
+ @discussion
+ Defaults to YES.
+ */
+@property (atomic) BOOL allowSocketConnection;
```
alex-hofsteede: Is there any way we can unify this and the `DISABLE_MIXPANEL_AB_DESIGNER` macro. 

We use that macro to effectively disable the designer connection by not adding the 4-fingers gesture recognizer. My concern is that now we have 4 possible combinations of this flag and the macro to think about. 

The reason it's a macro is that I figured that would be easier for people to add it to only their production build target.

alex-hofsteede:
OK the retry stuff looks ok, had one comment about the flag you added. otherwise looks good.

Mixpanel/Mixpanel.h
malectro: oh, yeah i missed that flag. can probably just use that.

malectro: 8b6e4ed

acanyon:
Thanks kyle! Do you think we could group a couple more of these pulls into the next release? eg. https://github.com/mixpanel/mixpanel-iphone/pull/218/files FYI. the build is failing b/c its a non-mixpanel pull.

And im updating the documentation anyway, so ill be happy to add a blurb about this feature in the ios docs. :)

Mixpanel/MPABTestDesignerConnection.m
```diff
@@ -218,21 +222,24 @@ - (void)webSocket:(MPWebSocket *)webSocket didCloseWithCode:(NSInteger)code reas
...
 {
     static int retries = 0;
-    if (self.sessionEnded || _connected || retries >= 10) {
+    
+    if (self.sessionEnded || _connected || retries >= maxRetries) {
         // If we deliberately closed the connection, or are already connected
         // or we tried too many times, then stop retrying.
         retries = 0;
-    } else if(initiate ^ (retries > 0)) {
+    } else if (initiate ^ (retries > 0)) {
```
joeatwork: This is just me being a tourist, but what's the deal with the caret operator here? I want to read this as `bitwise xor initiate and (retries > 0)` - is it the case that YES and NO are bitwise compliments, so we only run the block if exactly one of these is YES and the other is NO? Or does the binary caret operator mean other stuff in Objective C?

alex-hofsteede: @joeatwork  great question. yep, that's a XOR, the reasoning is this:

There are 3 cases.
`initiate = YES && (retries > 0) == NO`: This means you were already happily using the socket connection and it broke because the socket died somehow. That's when we try and get reconnected by calling`[socket open]` and also throwing a retry into the future in case this `[socket open]` doesn't work.

`initiate = NO && (retries > 0) == YES`: This is when we are in an exponential backoff retry loop, and somehow the first `[socket open]` didn't work. (ie `_connected` is still NO), so let's try again. (and kick off another retry even further in the future in case this one doesn't work either. 

`initiate = YES && (retries > 0) == YES`: This is the case where our `[socket open]` kindof worked but didn't go well (probably because the switchboard immediately rejected the connection for some reason) and the `webSocket:didFailWithError:` or `webSocket:didCloseWithCode` handler calls `reconnect` again because of that. In this case we are already in a retry backoff loop, so we don't actually want to start another one (because we know there is a queued up call to `reconnect` already scheduled for the future, or the maximum number of retries has already been reached) 

So yeah, TL;DR, we want to retry if we are either starting a reconnect cycle, or if we are already in one, but we don't want to start another cycle if we are already in one.

alex-hofsteede: Actually the 4th case, where both are NO is also kindof important. It means we successfully reconnected, or the max retries was reached, and we want to stop the reconnect loop.

joeatwork: Very cool! That's nicely expressed! (And thanks for the explanation!)

malectro:
@acanyon yeah, i'll be adding that pr to the release

PR #218, opened 2015-01-23
Add experimentsLoadedCallback.

DawnWright:
Hello!

For properly analyzing our retention data w.r.t. experiments, we need to track certain events only after we are certain that experiments have been successfully started. Mixpanel has no way of letting us know that experiments have been started, however. So I've added an optional callback to joinExperiments. We would like this to be part of the official SDK, and would appreciate getting your thoughts on this.

### DIFF
--- a/Mixpanel/Mixpanel.h
+++ b/Mixpanel/Mixpanel.h
@@ -634,6 +634,7 @@
  You do not need to call this method on the main thread.
  */
 - (void)joinExperiments;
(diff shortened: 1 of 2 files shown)


malectro:
thanks for this! we were talking about adding a callback so i think it'll be easy to add this to the next release.

DawnWright:
Excellent! We're very happy to see this will be happening for next release. Thank you!

PR #220, opened 2015-01-28
Cut down console spam

jadar:
It looks like some MPLog statements were left in the MPWebSocket class. I'm assuming they aren't supposed to be there. This PR changes them to MixpanelError and MixpanelDebug for errors and debug logs, respectively.

It's pretty annoying to have Mixpanel logs in my console when I'm trying to debug my app. 

Please let me know if I should change some of the MixpanelError macros over to MixpanelDebug. I'd be happy to do so.

### DIFF
--- a/Mixpanel/MPWebSocket.m
+++ b/Mixpanel/MPWebSocket.m
@@ -456,7 +456,7 @@ static __strong NSData *CRLFCRLF;
     NSInteger responseCode = CFHTTPMessageGetResponseStatusCode(_receivedHTTPHeaders);
 
     if (responseCode >= 400) {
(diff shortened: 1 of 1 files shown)


vvit:
Meanwhile you can use this solution: https://github.com/mixpanel/mixpanel-iphone/issues/221#issuecomment-72007969

welshm:
Would love to see this get pulled in. The spam in the console is fairly annoying

jadar:
I'd rather not use a workaround like that, it's messy. I just downgraded to 2.6.2.

malectro:
added to the latest release v2.7.1

jadar:
Glad to help!

PR #223, opened 2015-01-29
Change MPLog to MixpanelLogger in MPWebSocket

welshm:
MPLogger.h is already imported by MPWebSocket.m

Remove the declaration of MPLog function and use the MPLogger macros


malectro:
thanks for this @welshm. i've added it to the latest release.

PR #227, opened 2015-02-04
added a flushWithCompletion handler, closes #226

### LINKED ISSUE #226: Calling Mixpanel flush from application:didReceiveRemoteNotification:fetchCompletionHandler:
I want to log an event in this method, even if it is called while the application is in the background state. Currently, just using `trackEvent:` would just save the event, which would then be sent next time the application is brought to the foreground. 

Calling `flush` right after `trackEvent` Doesn't seem particularly safe though, since all flush does it place a task on a serial queue. Is flush supposed to be able to be called from the background state?

malectro:


acanyon:
:dancer: :sparkles:

maxstoller:
Just wondering if you have any sense when this might be merged into master? I'll use the branch for now. Thanks! :+1:

alex-hofsteede:
LGTM

PR #238, opened 2015-03-14
add an Xcode static library project file

morellid:
I don't use cocoa pods, I prefer using submodules, but I don't find the "copy the files from a repository" satisfactory, as I would lose track of the changes made on the mix panel subproject, and everything would look like a local change. This problem can be fixed simply having an Xcode static library project file, and importing mix panel as a subproject. This way in my repo I can just add a submodule, include the project, and the project will compile happily while mix panel files are (correctly) being tracked as a submodule, and not as main project files.

The only thing that is needed apart from dragging and dropping the project file into the app's project file is to link libicucore and add the mixpanel submodule location to the header search path.

### DIFF
--- a/mixpanel.xcodeproj/project.pbxproj
+++ b/mixpanel.xcodeproj/project.pbxproj
@@ -582,7 +582,7 @@
 				GCC_WARN_UNUSED_VARIABLE = YES;
 				IPHONEOS_DEPLOYMENT_TARGET = 8.2;
 				MTL_ENABLE_DEBUG_INFO = YES;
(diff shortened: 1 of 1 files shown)


davidplans:
Hey mixpanel peeps. We're the guys using mixpanel for Delta Rail (UK rail company) and for biobeats.com...having mixpanel as updatable xcode project instead of static folder would be awesome...so we thought we'd help out!

samgreen:
This looks like a duplicate of #252 so I'm going to close it, since #252 would resolve this issue as well as allowing building with Carthage.

PR #239, opened 2015-03-20
Add targeting to variant

zyskowsk:
add a targeting flag to variant.  Only merge people profiles if variant is from an experiment that is targeting

### DIFF
--- a/HelloMixpanel/ABTestingTests/test_decide_response.json
+++ b/HelloMixpanel/ABTestingTests/test_decide_response.json
@@ -15,7 +15,8 @@
             ],
             "tweaks":[],
             "id":1,
-            "experiment_id":1
+            "experiment_id":1,
+            "targeting": false,
         },
         {
             "actions":[
@@ -30,7 +31,8 @@
             ],
             "tweaks":[],
             "id":2,
-            "experiment_id":2
+            "experiment_id":2,
+            "targeting": false,
         }
     ]
 }
--- a/HelloMixpanel/ABTestingTests/test_decide_response.json
+++ b/HelloMixpanel/ABTestingTests/test_decide_response.json
@@ -16,7 +16,7 @@
             "tweaks":[],
             "id":1,
             "experiment_id":1,
-            "targeting": false,
+            "targeting": false
         },
         {
             "actions":[
@@ -32,7 +32,7 @@
             "tweaks":[],
             "id":2,
             "experiment_id":2,
-            "targeting": false,
+            "targeting": false
         }
     ]
 }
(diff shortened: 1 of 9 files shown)


HelloMixpanel/ABTestingTests/test_decide_response.json
```diff
@@ -15,7 +15,8 @@
             ],
             "tweaks":[],
             "id":1,
-            "experiment_id":1
+            "experiment_id":1,
+            "targeting": false,
```
zakj: no trailing commas in JSON, sadly

HelloMixpanel/ABTestingTests/test_decide_response.json
```diff
@@ -30,7 +31,8 @@
             ],
             "tweaks":[],
             "id":2,
-            "experiment_id":2
+            "experiment_id":2,
+            "targeting": false,
```
zakj: same

Mixpanel/MPVariant.m
```diff
@@ -137,6 +141,7 @@ - (id)initWithCoder:(NSCoder *)aDecoder
         self.experimentID = [(NSNumber *)[aDecoder decodeObjectForKey:@"experimentID"] unsignedLongValue];
         self.actions = [aDecoder decodeObjectForKey:@"actions"];
         self.tweaks = [aDecoder decodeObjectForKey:@"tweaks"];
+        self.targeting = [aDecoder decodeBoolForKey:@"targeting"];
```
alex-hofsteede: What happens with a library that is expecting this key and it isn't there?

zyskowsk: when would this happen?

Mixpanel/MPVariant.m
```diff
@@ -102,24 +102,28 @@ + (MPVariant *)variantWithJSONObject:(NSDictionary *)object {
         return nil;
     }
 
+    BOOL targeting = [[object objectForKey:@"targeting"] boolValue];
```
alex-hofsteede: What if this key doesn't exist. As a rule we are pretty strict about validation in these initializer class methods, so that we don't have to do any validation in the constructor.

Mixpanel/MPVariant.m
alex-hofsteede: sorry I actually left this comment in the wrong place. see above

alex-hofsteede: I thought it wouldn't but actually this is a valid concern. If an app saves an archive and then gets updated this code will be executed on the old archive which won't have the targeting property.

PR #246, opened 2015-04-09
Remove warnings generated with XCode 6.3 iOS 8.3 SDK

welshm:
- Rename overriding properties and create as readonly. Implement readonly property methods to check against the desired superclass property and return if appropriate.

### DIFF
--- a/Mixpanel.podspec
+++ b/Mixpanel.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = 'Mixpanel'
-  s.version      = '2.7.2'
+  s.version      = '2.7.3'
   s.summary      = 'iPhone tracking library for Mixpanel Analytics'
   s.homepage     = 'https://mixpanel.com'
   s.license      = 'Apache License, Version 2.0'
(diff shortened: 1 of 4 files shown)


Quuxplusone:
Please rebase and squash this patch into one or two commits (one commit per thing you're trying to fix). And 4b25cb4 just looks wrong to me; surely there's a better way to express the idea that e.g. the type of property `question` is covariant with the class type.

http://stackoverflow.com/questions/11418205/re-casting-a-property-in-subclasses looks relevant... but then, it also looks like what Mixpanel's code already does. What kind of compiler warning are you seeing? (You should cut and paste the warning(s) into your commit message(s) anyway, so that people looking at the history can figure out why you changed the code. Otherwise, a few months from now, someone will just "refactor" back to the simpler code, because there's nothing in the code nor comments that explains why the code suddenly got so complicated.)

welshm:
Going to create a new request - I've never tried to rebase/squash a patch before, and this seems to be in a mess now from the rebasing I've done.

Quuxplusone:
@welshm FYI, you just "squashed" one too many revisions. You should have "picked" your first change, and then "squashed" the subsequent ones into that one; instead you "squashed" all your changes into @alex-hofsteede's.

You could fix it by doing `git fetch; git rebase -i origin/master` if you wanted to.

welshm:
@Quuxplusone Thanks for the pointer! I've created a new pull request and recreated my fork as well as it got into a pretty bad state.

PR #251, opened 2015-04-21
Making sure data is archived even if the host app has crashed

zats:
Hi, we noticed that some of our users don't have any super-properties
After investigation, here is a scenario that caused the problem:
Host application crashes, mixpanel looses all the data since it keeps all the properties in memory in-between app being backgrounded / terminated.

A possible solution: not to remove the archived plist files after reading them on startup but it would only preserve last serialised state not the latest set one, which is not desirable when dealing with sensitive user data.

Proposed solution: become one of the chain of the exception handlers and archive data upon exception

### DIFF
--- a/Mixpanel/Mixpanel.m
+++ b/Mixpanel/Mixpanel.m
@@ -30,6 +30,15 @@
 
 #define VERSION @"2.7.3"
 
+static NSHashTable *mp_mixpanelInstances;
+static NSUncaughtExceptionHandler *mp_originalExceptionHandler;
+static void mp_uncaughtExceptionHandler(NSException *exception) {
+    for (Mixpanel *mixpanel in mp_mixpanelInstances) {
+        [mixpanel archive];
+    }
+    mp_originalExceptionHandler(exception);
+}
+
 @interface Mixpanel () <UIAlertViewDelegate, MPSurveyNavigationControllerDelegate, MPNotificationViewControllerDelegate> {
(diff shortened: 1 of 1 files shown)


zats:
hey @alex-hofsteede! Can you review this one please? It's a short one I promise :smile_cat:

cmcewen:
:+1:

cmcewen:
@zats are you using this in production? seems like we may be missing a few events and wondering if this is the cause

zats:
@cmcewen we are using a slightly modified version since we didn't want to change the source.
But we actually got an email from mixpanel saying that proposed solution is not production safe (without any technical details) and they won't probably merge it. 
We're still relying on this approach, but you should use your best judgment.
I'd love you if or anyone for that matter can suggest reasons why this code is not production-safe

samgreen:
We're working on a modified version internally that will resolve this. Just need to finalize some details with the folks at Crashlytics.

zats:
great, I'm looking forward to it! if there is a beta version, I'd love to test it :+1:

samgreen:
@zats Sure thing, I'll reach out next week.

samgreen:
This will be resolved with #306

uzysjung:
Hi, Actually It's not resovled.
It should be use signal handler not a exception handler.

exception handler can handle specific situations such as segmentation fault, invalid instruction, system abort. 

plz reopen

Issue #207, opened 2014-12-24
signal SIGABRT crash at showTakeoverNotificationWithObject


clkr:
UIStoryboard *storyboard = [UIStoryboard storyboardWithName:@"MPNotification" bundle:nil]; crashes with signal SIGABRT when there is a takeover notification.

Function:
- (BOOL)showTakeoverNotificationWithObject:(MPNotification *)notification

arittr:
@clkr thanks for this - i'm going to see if we can patch this edge case going forward, but for now you might want to just copy over the XIB/resource manually into your app bundle. i believe this is happening because our library assumes the storyboard to be in the app bundle, which isn't always the case - let me know if the manual copy helps at all!

clkr:
@arittr copying manually solves the crash, thanks! I've also noticed that Titanium guys had a similar problem and found a solution, it might help:
https://github.com/hyperlab/TiMixpanel/issues/4

arittr:
@clkr fantastic - so happy we got this fixed. thanks so much for sending this my way - great solution and very well documented. will be passing this around until we patch the lib to avoid the issue. thanks!

clkr:
@arittr I'm happy to hear this helps. Cheers!

fluidsonic:
Same issue here.

This issue occurs because Mixpanel assumes that its storyboard files are always located in the app's main bundle. This is not the case with the new framework structure introduced with Xcode 6 adopted by CocoaPods 0.36.

In `Mixpanel.m` the storyboard is loaded from the app's main bundle instead of from the bundle where the Mixpanel library is located.
```objc
[UIStoryboard storyboardWithName:@"MPNotification" bundle:nil];
```
should be
```objc
[UIStoryboard storyboardWithName:@"MPNotification" bundle:[NSBundle bundleForClass:Mixpanel.class]];
```

See http://blog.cocoapods.org/CocoaPods-0.36/ (section 'Caveats')
See http://blog.cocoapods.org/Pod-Authors-Guide-to-CocoaPods-Frameworks/

This issue also affects other parts of the Mixpanel library like the survey.

clkr:
@fluidsonic it will be useful to create a pull request if this fixes the issue.

fluidsonic:
@clkr yeah already on it :)

fluidsonic:
Fixed in PR #250

PR #252, opened 2015-04-22
Updated mixpanel to build itself with Carthage.

simX:
This allows Mixpanel to be built using Carthage: https://github.com/Carthage/Carthage .  Note: this change was originally created by GitHub user justinmakaila, but in my testing it seems to work fine.


Mixpanel/MPObjectSelector.h
```diff
@@ -6,7 +6,7 @@
 //  Copyright (c) 2014 Mixpanel. All rights reserved.
 //
 
-#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
```
Quuxplusone: This header file doesn't use anything from UIKit; Foundation should be sufficient. (You might need to add UIKit somewhere else, though.)

simX:
Updated the pull request to fix quuxplusone's comment, and to merge in the latest Mixpanel changes.

simX:
Also, the automatic Travis CI build continually gives me an error here on GitHub:

The command "openssl aes-256-cbc -k "$ENCRYPTION_SECRET" -in scripts/profile/HelloMixpanel_Ad_Hoc.mobileprovision.enc -d -a -out scripts/profile/HelloMixpanel_Ad_Hoc.mobileprovision" failed and exited with 1 during .

Not sure what that's about.

pcantrell:
+1 Works for me
Fixes #265

simX:
Actually this PR has a bug in the way that the framework looks for xib files when presenting a survey that will cause it to crash.  I have an updated fork that will fix this problem.

samgreen:
I really want to get Mixpanel compatible with Carthage! This is definitely a priority for us. 

That said, I think providing two separate `.xcodeproj` files in the same repo might be kind of confusing for users who simply clone master. I'm researching Carthage now to see if there's an alternative to this.

Do you guys have any thoughts as to making this process as simple and clear as possible for both CocoaPods and Carthage users?

simX:
We can put the framework target in the existing HelloMixpanel.xcodeproj file, but it didn't seem appropriate because it was just sample code; there weren't any targets for building the static library.  Also, there are some quirks about finding the xcodeproj file that makes it easier to exist at top level.

samgreen:
Okay, we're gonna go ahead and move forward with this PR. I've verified it locally and it seems to building well with Carthage. @simX Can you please address the `UIKit` import comment above, and ensure we're **not** importing `UIKit` anywhere that it is not needed? After that I'll feel comfortable merging these changes in to master.

simX:
I've merged in the latest Mixpanel changes.  I've tidied up the project file a bit, fixed some build errors due to removal of the libMPCategoryHelpers.a static library, and fixed a build warning.  I've also reverted all the unnecessary imports of UIKit.

simX:
Note: going forward, this new Xcode project file will need to be kept up-to-date with the latest file additions and removals as they happen (such as the removal of the libMPCategoryHelpers.a static library).  An xcodebuild or carthage build might need to be added to any battery of tests before each release.

simX:
I looked in issue #265 and noticed @Rupert-RR's comment about changing references to the mainBundle.  A similar issue was fixed in #207.  I've noticed a number of other places where mainBundle is referenced.  Should we change those to explicitly use bundleForClass: instead of mainBundle as well?

samgreen:
@simX The bundle stuff needs to be fixed everywhere. I thought we were using `bundleForClass:` everywhere in the code base. Let me know what you find.

simX:
@samgreen Just took a look, it seems I was wrong.  The framework uses [NSBundle mainBundle] in some places, but it's appropriate usage.  In various places you're getting the CFBundleVersion, CFBundleShortVersionString, and CFBundleIdentifier from the main application for super-properties and people properties, as well as supported interface orientations and snapshot_config.json, which all seem to appropriately come from the main bundle.

I think this pull request is ready. :)

samgreen:
@simX Thank you so much for this contribution. Merging now.

alec-c4:
great, but could you please update readme?

PR #257, opened 2015-05-03
removed 2 NSLog() calls

adamwulf:
Changed the last 2 NSLog calls to MixpanelError calls


samgreen:
I don't see any issue merging this. Paging @alex-hofsteede

Issue #232, opened 2015-02-19
`reset` should reset `shownSurveyCollections` and `shownNotifications`


peymano:
In our app, when a user signs out, we call Mixpanel's `reset` method. If a different user signs in, user 2 wont see any in-app notifications that user 1 has already seen.

This seems like an oversight of the `reset` method. It resets all archived properties except `self.shownSurveyCollections` and `self.shownNotifications`.

```
- (void)reset
{
    dispatch_async(self.serialQueue, ^{
        self.distinctId = [self defaultDistinctId];
        self.nameTag = nil;
        self.superProperties = [NSMutableDictionary dictionary];
        self.people.distinctId = nil;
        self.people.unidentifiedQueue = [NSMutableArray array];
        self.eventsQueue = [NSMutableArray array];
        self.peopleQueue = [NSMutableArray array];
        self.timedEvents = [NSMutableDictionary dictionary];
        [self archive];
    });
}
```

Grubas7:
Same with surveys.

Grubas7:
Probably in reset method, boolean value of `decideResponseCached` should be set to NO to allows `-checkForDecideResponseWithCompletion: useCache:` method call success block of `if (!useCache || !self.decideResponseCached)`.

samgreen:
This is resolved in #295

yas375:
@Grubas7 @samgreen thanks for taking care of this!

Re: 3b31d3f8a3dfb6adf95b0adaf23b96e756525680
Could you please elaborate a little more why `self.shownNotifications` doesn't need to be reset too as it was suggested by @peymano?

yas375:
In my understanding `self.shownNotifications` needs to be reset too:


```
- (void)checkForDecideResponseWithCompletion:(void (^)(NSArray *surveys, NSArray *notifications, NSSet *variants, NSSet *eventBindings))completion useCache:(BOOL)useCache
{
    dispatch_async(self.serialQueue, ^{
        MixpanelDebug(@"%@ decide check started", self);
        ...
        if (!useCache || !self.decideResponseCached) {
          ...
        }
        ...
        NSArray *unseenNotifications = [self.notifications objectsAtIndexes:[self.notifications indexesOfObjectsPassingTest:^BOOL(id obj, NSUInteger idx, BOOL *stop) {
            return [self.shownNotifications member:@(((MPNotification *)obj).ID)] == nil;
        }]];
        ...
    });
}
```

[...]

PR #260, opened 2015-05-07
Clear variants and shownNotifications variable in reset function

jordanmnunez:
@alex-hofsteede 

In a kiosk situation, apps will call mixpanel.reset() at the end of every anonymous session. Currently, past variants are persisting through this reset. This is causing no new experiments to be displayed once one of each variant has been received. By including variants and shownNotifications in the reset we can allow AB tests to show the most recent variant that has been received from decide.

### DIFF
--- a/Mixpanel/Mixpanel.m
+++ b/Mixpanel/Mixpanel.m
@@ -537,6 +537,9 @@ static __unused NSString *MPURLEncode(NSString *s)
         self.eventsQueue = [NSMutableArray array];
         self.peopleQueue = [NSMutableArray array];
         self.timedEvents = [NSMutableDictionary dictionary];
+        self.variants = [NSSet set];
+        self.shownNotifications = [NSMutableSet set];
+        
         [self archive];
     });
 }


jordanmnunez:
@alex-hofsteede any opinions as to whether this is a viable solution?

alex-hofsteede:
@jordanmnunez not sure this is enough. If there are variants running, we would need to cleanly stop them before just resetting `self.variants`, as they add all kinds of hooks into the UI code.

Grubas7:
Problem also occurs in surveys.
1. Login as a some user (eg. UserX) and set `MixpanelIdentify`
2. Show and answer for some survey
3. Logout (via method reset)
4. Login as a other user (eg. UserY) and set `MixpanelIdentify`
5. Last survey (which seen only UserX) is marked as already answered

Probably array `shownSurveyCollections` need to be released, and bool `decideResponseCached` reset to NO.

arittr:
@jordanmnunez @alex-hofsteede new issue about this today : https://github.com/mixpanel/mixpanel-iphone/issues/232#issuecomment-102010207

samgreen:
This should be resolved in #295

jordanmnunez:
<3

Issue #261, opened 2015-05-08
Crash in MPWebSocket.m:1277 (call to assertOnWorkQueue)


dgatwood:
I just saw a fascinating crash report in MixPanel code that occurred only about 1.5 seconds after our app launched.  Apparently, assertOnWorkQueue failed even though the call path very clearly showed that the code was running within that work queue....

The only thing I can imagine is that the MPWebSocket object was getting deallocated while the code was running (ick) and therefore the call to dispatch_get_specific was comparing a newly nil self pointer with an arbitrary void pointer value from the dispatch queue that wouldn't have gotten zeroed.  If I'm right, then this is probably just another manifestation of https://github.com/mixpanel/mixpanel-iphone/issues/234 caused by the MPWebSocket object going away while other code is still actively using it.  However, I can't be certain, hence I'm filing a separate ticket.

The backtrace is shown below.

<pre>
Thread : Crashed: Thread
0  libsystem_kernel.dylib         0x0000000197a4f270 __pthread_kill + 8
1  libsystem_pthread.dylib        0x0000000197aed170 pthread_kill + 112
2  libsystem_c.dylib              0x00000001979c6b18 abort + 112
3  libsystem_c.dylib              0x000000019799ba3c basename
4  HotspotVPN                     0x000000010012a434 -[MPWebSocket assertOnWorkQueue] (MPWebSocket.m:368)
5  HotspotVPN                     0x0000000100132154 -[MPWebSocket _pumpScanner] (MPWebSocket.m:1277)
6  HotspotVPN                     0x00000001001332a8 __34-[MPWebSocket stream:handleEvent:]_block_invoke523 (MPWebSocket.m:1432)
7  libdispatch.dylib              0x0000000197909994 _dispatch_call_block_and_release + 24
8  libdispatch.dylib              0x0000000197909954 _dispatch_client_callout + 16
9  libdispatch.dylib              0x00000001979140a4 _dispatch_queue_drain + 1448
10 libdispatch.dylib              0x000000019790ca5c _dispatch_queue_invoke + 132
11 libdispatch.dylib              0x0000000197916318 _dispatch_root_queue_drain + 720
12 libdispatch.dylib              0x0000000197917c4c _dispatch_worker_thread3 + 108
13 libsystem_pthread.dylib        0x0000000197ae922c _pthread_wqthread + 816
</pre>

samgreen:
Waiting on #262 review

samgreen:
[...]

Issue #234, opened 2015-02-28
Crash in Websocket A/B in the iOS Simulator


joeatwork:
From a customer report

Starting a couple of days ago, I am getting very frequent crashes in MPWebSocket.m:1714 (iOS SDK) when running in the simulator with iOS version 7.0.X. The crash happens within a minute or so of running the app regardless of what I am doing. We have not seen this crash in the wild to date.

The crash is EXC_BAD_ACCESS, but the run loop (_runLoop) looks fine, so it's not clear what object was actually deallocated.

Here is the stack trace on the problem thread:

```
* thread #10: tid = 0xc53246, 0x0000000110207e3d libsystem_platform.dylib`_spin_lock + 7, stop reason = EXC_BAD_ACCESS (code=1, address=0x14)
frame #0: 0x0000000110207e3d libsystem_platform.dylib`_spin_lock + 7
frame #1: 0x000000010fbf19d1 CoreFoundation`CFSocketEnableCallBacks + 49
frame #2: 0x000000010c35ff98 CFNetwork`SocketStream::securityBufferedRead_NoLock() + 356
frame #3: 0x000000010c35fde1 CFNetwork`SocketStream::socketCallbackReadLocked(SocketStreamSignalHolder*) + 99
frame #4: 0x000000010c35d997 CFNetwork`SocketStream::socketCallback(__CFSocket*, unsigned long, __CFData const*, void const*) + 139
frame #5: 0x000000010c35d8de CFNetwork`SocketStream::_SocketCallBack_stream(__CFSocket*, unsigned long, __CFData const*, void const*, void*) + 64
frame #6: 0x000000010fbb13e0 CoreFoundation`__CFSocketPerformV0 + 656
frame #7: 0x000000010fb7cec1 CoreFoundation`__CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE0_PERFORM_FUNCTION__ + 17
frame #8: 0x000000010fb7c792 CoreFoundation`__CFRunLoopDoSources0 + 242
frame #9: 0x000000010fb9861f CoreFoundation`__CFRunLoopRun + 767
frame #10: 0x000000010fb97f33 CoreFoundation`CFRunLoopRunSpecific + 467
frame #11: 0x000000010d1a7b1e Foundation`-[NSRunLoop(NSRunLoop) runMode:beforeDate:] + 253
[...]

PR #262, opened 2015-05-08
Possible fix for issue 261 and issue 234.

dgatwood:
Basic changes:

1.  Consolidated the socket cleanup path so that each location does everything.

2.  Switched from using the self pointer (which might in weird circumstances get nil'ed out while the code is running) to using the queue itself in setspecific/getspecific in an attempt to fix https://github.com/mixpanel/mixpanel-iphone/issues/261.

3.  Significantly reduced the number of places where the self-referencing pointer is set to nil, to ensure that it doesn't happen until any code that depends on "self" has finished.

Thus far, with these changes, I haven't seen the weird crashes in https://github.com/mixpanel/mixpanel-iphone/issues/234 (_MPRunLoopThread main crash in CFSocketEnableCallBacks/_spin_lock).  However, the bug was nondeterministic, so this isn't guaranteed to fix the problem.

### DIFF
--- a/Mixpanel/MPLogger.h
+++ b/Mixpanel/MPLogger.h
@@ -11,6 +11,10 @@
 #ifndef MPLogger_h
 #define MPLogger_h
 
+#if DEBUG
+#define MIXPANEL_DEBUG
+#endif
+
 static inline void MPLog(NSString *format, ...) {
     __block va_list arg_list;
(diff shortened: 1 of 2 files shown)


samgreen:
This is resolved with #302 and will be out with the next release.

Issue #270, opened 2015-06-03
EXC_BAD_ACCESS in in MPApplicationStateSerializer.m


weibel:
line 41: screenshotImageForWindowAtIndex: I often get a EXC_BAD_ACCESS  soon after initialization on
```[window drawViewHierarchyInRect:window.bounds afterScreenUpdates:NO]```

The log says
> <Error>: CGContextSaveGState: invalid context 0x0. This is a serious error. This application, or a 
> library it uses, is using an invalid context  and is thereby contributing to an overall degradation of 
> system stability and reliability. This notice is a courtesy: please fix this problem. It will become a fatal 
> error in an upcoming update.

Any ideas?

**Comment by weibel** [comment]
I think I got the error. The details of the window at the time of the crash are

```
<UIWindow: 0x156f98c20; frame = (0 0; 0 0); clipsToBounds = YES; hidden = YES; 
userInteractionEnabled = NO; gestureRecognizers = <NSArray: 0x171649d80>; 
layer = <UIWindowLayer: 0x171229660>>
```

It's odd the frame is CGRectZero.

samgreen:
#271 should resolve this

samgreen:
Closing as resolved.

PR #271, opened 2015-06-05
Prevent crashes when devices are connected to A/B tests

weibel:
Fix for https://github.com/mixpanel/mixpanel-iphone/issues/270


arittr:
thanks @weibel for the PRQ - we really appreciate it! @alex-hofsteede can you take a look at this and #270?

samgreen:
Looks good, thanks!

PR #272, opened 2015-06-15
added support for the $ignore_time flag setting

mishagray:
I want to be able to set/increment properties without updating the "Last Seen" flag in the Profiles.  (Useful for tracking background tasks, like GeoFencing, etc).

### DIFF
--- a/Mixpanel/Mixpanel.h
+++ b/Mixpanel/Mixpanel.h
@@ -692,6 +692,21 @@
  People</b>.
  */
 @interface MixpanelPeople : NSObject
+/*!
+ @property
+ 
+ @abstract
+ controls the $ignore_time property in any subsequent MixpanelPeople operation.
+ 
+ If the $ignore_time property is present and true in your request,
+ Mixpanel will not automatically update the "Last Seen" property of the profile.
+ Otherwise, Mixpanel will add a "Last Seen" property associated with the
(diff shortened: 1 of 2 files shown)


arittr:
@mishagray thanks so much for this, really appreciate the help - we'll take a look at merging this into our next release.

samgreen:
@mishagray I'm a bit confused by this PR. Couldn't you just set `$ignore_time`manually for a user based on when you want this behavior?

mishagray:
So the primary reason is it makes it 100% easy to deal with all the events
in your app that get triggered when it starts up, but is starting due to a
background event (like a geofence tigger, or a background refresh).

iOS will even re-recreate your root view controller, which can receive a
ViewDidAppear (even though it still in the background).

It's actually HARD to deal with this on an event-by-even basis, as your app
gets more complicated, since we have analytics around API events,
ViewDidAppears, etc.   And we WANT to see the analytics for things
happening in the background, but we don't want the MixPanel 'people' logic
to show these users as 'seen', since this is really just some automatic
logic running in the background on their phones.

I just put logic to set '$ignore_time" when the application state changes.
So any events that happen in a background state are still recorded, but
don't show the 'user' as being active.



- michael.gray@insertwonderhere.com

samgreen:
We had some conflicts in this, so I resolved them in #401. Kept your commits though.

PR #274, opened 2015-06-16
fix Xcode7 warning: Conflicting return type in implementation of 'supportedInterfaceOrientations'

Coeur:


Coeur:
Hello @samgreen,

Why did you revert this change with your commit 1d379d4?

samgreen:
@Coeur Because this change is not correct. While it is more pragmatic, it specifies a different return type than the super class. Check the docs here: https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIViewController_Class/#//apple_ref/occ/instm/UIViewController/supportedInterfaceOrientations

Coeur:
Well, UIViewController.h says:

    @interface UIViewController (UIViewControllerRotation)
    
    // New Autorotation support.
    - (BOOL)shouldAutorotate NS_AVAILABLE_IOS(6_0);
    - (UIInterfaceOrientationMask)supportedInterfaceOrientations NS_AVAILABLE_IOS(6_0);

So one day Apple will update its online doc I suppose and we will update the signature of this method to finally fix the warning. It wasn't incorrect to do it now, as NSUInteger and UIInterfaceOrientationMask are compatible.

samgreen:
I show this in Xcode 6 (iOS 8.4). If you're using Xcode beta, keep in mind we don't support that.

    - (BOOL)shouldAutorotate NS_AVAILABLE_IOS(6_0);
    - (NSUInteger)supportedInterfaceOrientations NS_AVAILABLE_IOS(6_0);

Coeur:
Well, *we* also has the meaning of the clients of mixpanel. And they might be using Xcode 6 using strict compilation options, or they might be using Xcode 7 using strict compilation options as well. Maybe we could deactivate the clang warning for this specific line while transitioning from Xcode 6 to Xcode 7. 

    // removing warning for Xcode 6 and earlier
    #pragma clang diagnostic push
    #pragma clang diagnostic ignored "-Wall"
    - (UIInterfaceOrientationMask)supportedInterfaceOrientations {
    #pragma clang diagnostic pop

(I know it is not "-Wall", but not sure which one it is exactly, as I only have El Capitan, so I can't install Xcode 6 anymore)

PR #280, opened 2015-07-06
Remove bare PNGs in favor of Xcode image sets. 

### LINKED ISSUE #275: Duplicate asset files are included in the podspec
The podspec has the following:

```ruby
ss.resources 	 = ['Mixpanel/**/*.{png,storyboard}']
```

Which causes some .png files inside the Mixpanel directory to be included, along with files with the same name within the `.xcasset` file which caused a build warning when compiled as a framework.

- https://github.com/mixpanel/mixpanel-iphone/blob/master/Mixpanel/MPCloseBtn.png
- https://github.com/mixpanel/mixpanel-iphone/blob/master/Mixpanel/Media.xcassets/MPCloseBtn.imageset/MPCloseBtn.png

screen shot 2015-06-16 at 11 48 44

samgreen:
Resolves #275


alex-hofsteede:
LGTM

Issue #263, opened 2015-05-14
MPObjectSelector calls view-related methods from background thread


mad102190:
In `MPObjectSelector.m`, class `MPObjectFilter`, the method `getChildrenOfObject: ofType:` (line 379) has the following check:
```objective-c
        if (!class || [((UIViewController *)obj).view isKindOfClass:class]) {
            [children addObject:((UIViewController *)obj).view];
        }
```
This is running on a background thread and can actually cause issues with app-wide animations, since calling `.view` on a UIViewController does more than simply access the view property. Instead, it actually tries to load the view in question, which shouldn't happen on the background thread. 

I've added code in my app that allows me to add a symbolic breakpoint on unsafe background calls (as described in a SO post [here](http://stackoverflow.com/a/26168956), and it gets called many times on app launch due to this call. My animations sometimes get disabled completely in the app (the OS sometimes does this automatically when unsafe calls are made in the background) due to this issue.

samgreen:
This will be resolved in v2.8.2

samgreen:
This should be resolved now if you want to do a `pod install`.

PR #281, opened 2015-07-07
Ensure MPObjectSelector does not accidently invoke loadView from a background thread

### LINKED ISSUE #231: Crash related to MPObjectSelector.m
MPObjectSelector.m line 378
-[MPObjectFilter getChildrenOfObject:ofType:]

causes our app to crash. It happens after our table view controller's viewDidLoad is called.

samgreen:
This should resolve #231 and #263.


alex-hofsteede:
LGTM

Issue #276, opened 2015-06-23
Crash on -[Mixpanel flushQueue:endpoint:]


vj9:
Fatal Exception: NSRangeException
*** -[__NSArrayM objectAtIndex:]: index 4 beyond bounds for empty array

0	CoreFoundation	__exceptionPreprocess
2	CoreFoundation	-[__NSArrayM removeObjectAtIndex:]
3     CoreFoundation	-[NSMutableArray removeObjectsInRange:inArray:range:]
4	Mixpanel.m line 775 -[Mixpanel flushQueue:endpoint:]
5 	Tracker.m line 132 -[Tracker flush]
6 	Tracker.m line 115 -[Tracker stop]
7 	PeopleMetricsTask.m line 68 -[PeopleMetricsTask perform]
8 	Task.m line 61 __19+[Task runClasses:]_block_invoke
9      Foundation	__NSBLOCKOPERATION_IS_CALLING_OUT_TO_A_BLOCK__
10    Foundation	-[NSBlockOperation main]
11	Foundation	-[__NSOperationInternal _start:]
19	libsystem_pthread.dylib	start_wqthread

arittr:
hey @vj9 - what version of the library is this? are you running the Mixpanel SDK on multiple threads? this looks to me like somehow an element in the queue to be flushed is being changed/removed during the flushQueue process.

vj9:
Using mixpanel 2.6.2. Yeah this is being called in a separate thread.

arittr:
hey @vj9 - our current release is v2.8.1 so you might want to try upgrading and see if that fixes the issue. that being said I think threading is the issue here - what operations are you doing from the separate thread? is there any possibility that this array is being touched from two threads at once (like tracking from one and flushing from the other)?

bkvirendra:
I'm using the 2.8.1 and still getting the same crash at [Mixpanel flushQueue:endpoint:]

PR #282, opened 2015-07-08
Synchronize flushQueue using each queue as the mutex token

samgreen:
Ensure the flushQueue method synchronizes on the queue parameter that is passed in. This should prevent issues with multi-threaded code that attempts the mutate the underlying data structures.

### DIFF
--- a/Mixpanel/Mixpanel.m
+++ b/Mixpanel/Mixpanel.m
@@ -616,34 +616,36 @@ static __unused NSString *MPURLEncode(NSString *s)
 
 - (void)flushQueue:(NSMutableArray *)queue endpoint:(NSString *)endpoint
 {
-    while ([queue count] > 0) {
-        NSUInteger batchSize = ([queue count] > 50) ? 50 : [queue count];
-        NSArray *batch = [queue subarrayWithRange:NSMakeRange(0, batchSize)];
+    @synchronized (queue) {
+        while ([queue count] > 0) {
+            NSUInteger batchSize = ([queue count] > 50) ? 50 : [queue count];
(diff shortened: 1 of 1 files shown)


samgreen:
This should resolve #276.

alex-hofsteede:
Can I get a rationale for this change? 

The way the curernt model works, we should always be mutating these lists on the serial queue, so race conditions should not be a problem. If there are other threads modifying the underlying data structures in a non safe manner, then we should fix that. If we can point to a specific case where it truly is best to be modifying the underlying data from another thread, then we can consider adding synchronized.

The main point is that if the code is doing something bad. I'd rather just stop doing the bad thing instead of adding another safety net.

samgreen:
My rationale for this change is to be more defensive in protecting the call to `flush`. I protected this code in response to #276. As it stands now, if users of our library call `flush` on a background thread, and our flush timer calls it on the main thread, we will have a race condition as both threads attempt to mutate the queues.

samgreen:
To be clear, this change is not due to any of the code we've written and is to protect consumers of our library. It may make more sense as an assert to ensure its only called from the main thread. In which case we just need to document to users this method is NOT thread safe.

alex-hofsteede:
Everything inside `[mixpanel flush]` is a block which is added to the serial queue. Access to these structures should already be 'serialized' by virtue of the serial queue. Obviously we're breaking that somewhere, and we need to fix that, but adding another layer of of serialization safety net doesn't fix that so much as cover up a bug that we have somewhere.

samgreen:
Ah I missed the higher level call. Got it.

PR #283, opened 2015-07-09
Automated conversion to modern syntax.

samgreen:
This is an automated conversion performed by Xcode. This will ensure we're using `instancetype` instead of `id` where appropriate. It also makes use of the `NS_DESIGNATED_INITIALIZER` macro to increase clarity and aid interoperability with swift.

### DIFF
--- a/HelloMixpanel/ABTestingTests/ABTestingTests.m
+++ b/HelloMixpanel/ABTestingTests/ABTestingTests.m
@@ -53,7 +53,7 @@
 @end
 @implementation A
 
-- (id)init
+- (instancetype)init
 {
     if((self = [super init])) {
         self.count = 0;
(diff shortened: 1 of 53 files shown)


joeatwork:
This is a question, not a suggestion - but does this change effect the versions of XCode/Build tools people can use to build the library? (Even if the answer is yes, it isn't necessarily relevant, we may have a base version we require that knows how to build this stuff already)

samgreen:
This change requires Xcode 5 to support the `instancetype` declaration. I should also note that Xcode 6 is required to build for iOS 8+. Based on that fact, I am assuming that the large majority of developers will have already upgraded to Xcode 6. If not, it's a simple click from the App Store.

alex-hofsteede:
LGTM.

Issue #241, opened 2015-03-26
Flush Timer not started on initWithToken:launchOptions:andFlushInterval:


originell:
The interval is set via `_flushInterval` , which skips the respective setter of course (which would in turn call `startFlushTimer`).

I noticed that because I'm working on a project where I don't have to start Mixpanel right in the AppDelegate, but much later on. So there's also no `applicationDidBecomeActive` to be called.
I wondered why nothing seemed to work, except when I flush myself. Then I discovered that simple fact :) 

Current workaround is to set the `flushInterval` again after the init, manually.

```
mixpanel = Mixpanel(token: MixpanelToken, launchOptions: nil, andFlushInterval: 60)
mixpanel.checkForNotificationsOnActive = true
mixpanel.showNotificationOnActive = false  // Show Notifications manually.
mixpanel.checkForSurveysOnActive = true
mixpanel.showSurveyOnActive = false  // Show Surveys manually.
mixpanel.flushInterval = 60  // Workaround: startFlushTimer not called on Mixpanel init
mixpanel.flushOnBackground = true
```

Initially wanted to do a PR that fixes that, but wasn't sure what would be your preferred way of solving this :) 

Cheers

uooq:
After a half-day of frustration and a long interaction with some smart Mixpanel support folks, I finally got enough debugging info to figure out that this was going on.

+1 on a please fix request!

Grubas7:
@originell
Thanks, I had same problem.

samgreen:
See my comment on #286

PR #286, opened 2015-07-13
Fixing lack of firing flush timer in -initWithToken:launchOptions:andFlushInterval: method.

Grubas7:
Fix for issue https://github.com/mixpanel/mixpanel-iphone/issues/241

### DIFF
--- a/Mixpanel/Mixpanel.m
+++ b/Mixpanel/Mixpanel.m
@@ -135,7 +135,7 @@ static Mixpanel *sharedInstance = nil;
     if (self = [self init]) {
         self.people = [[MixpanelPeople alloc] initWithMixpanel:self];
         self.apiToken = apiToken;
-        _flushInterval = flushInterval;
+        self.flushInterval = flushInterval;
         self.flushOnBackground = YES;
         self.showNetworkActivityIndicator = YES;
 


samgreen:
Hi @Grubas7, thank you so much for the pull request. We really appreciate when the community wants to give back. 

Unfortunately this change will alter the public facing API in a non intuitive way, and could possibly break code that is depending on existing behavior. We are currently discussing other options internally such as providing an additional constructor that will immediately start the flush timer.

Grubas7:
IMO right now work of SDK is non intuitive. I think @originell and @uooq would agree with me ;)
Why? When you create Mixpanel instance by `-initWithToken:andFlushInterval:` method there won't be any flush after time you could expect. And there's no public method to manually start/stop flushing timer. Only possible way to force SDK to periodically flush is set property `-flushInterval:` (might be even default value) after init.

In other words:
`-initWithToken:andFlushInterval:` doesn't start flushing timer
`-flushInterval:` starts flushing timer
Same parameter name, but different reaction.

alex-hofsteede:
@Grubas7 yeah we agree that the current behavior is not very intuitive, but we also need to evaluate whether changing it would break Mixpanel implementations for customers that are currently relying on the existing behavior, which has existed for some time. Pushing a behavior change like this is therefore not as trivial as it might seem. 

If we were to implement this change, we would also need to consider situations where e.g. a Mixpanel instance is being initialized during a background task, and whether we should still automatically start the flush timer in that case.

Grubas7:
@samgreen @alex-hofsteede As long as `-flushInterval:` property starts timer there's no problem for me ;) But maybe putting this issue to 'know issues' would be helpful for other devs.

samgreen:
I'm going to go ahead and clearly document the behavior for now. We can consider changing the internal behavior in the next major release.

originell:
Thanks for doing that! :)

PR #288, opened 2015-07-14
Move private source from static library in to master

samgreen:


Mixpanel/UIView+MPHelpers.m
```diff
@@ -0,0 +1,213 @@
...
+ 
+ Creates a short string which is a fingerprint of a UIButton's image property.
+ It does this by downsampling the image to 8x8 and then downsampling the resulting
+ 32bit pixel data to 8 bit. This should allow us to select images that are identical or
+ almost identical in appearance without having to compare the whole image.
+ 
+ Returns a base64 encoded string representing an 8x8 bitmap of 8 bit rgba data
+ (2 bits per component).
+ */
+- (NSString *)mp_varC
```
alex-hofsteede: can you give all these guys back their original names and just have a section at the bottom where you alias them eg. 

```
- (NSString *)mp_varC { return [self mp_imageFingerprint]; }
- (NSString *)mp_varD { ...
```

We're kinda stuck with these obfuscated names forever because we use them in A/B testing path selectors, but I'd prefer that the code at least become somewhat readable again.

HelloMixpanel/MPCategoryHelpersTests/MPCategoryHelpersTests.m
```diff
@@ -0,0 +1,84 @@
...
+- (void) testExistance
+{
+    UIView *v1 = [[UIView alloc] init];
+    XCTAssert([v1 respondsToSelector:@selector(mp_varA)]);
+    XCTAssert([v1 respondsToSelector:@selector(mp_varB)]);
+    XCTAssert([v1 respondsToSelector:@selector(mp_varC)]);
+    XCTAssert([v1 respondsToSelector:@selector(mp_varSetD)]);
+    XCTAssert([v1 respondsToSelector:@selector(mp_varE)]);
+    
+    XCTAssertFalse([v1 respondsToSelector:@selector(mp_nonexistant)]);
```
Quuxplusone: `s/xistan/xisten/g`

samgreen: Thanks!

Mixpanel/UIView+MPHelpers.m
samgreen: Added.

alex-hofsteede:
Actually 1 more request, can you refactor so that the `mp_varX` functions do the encrypting of the result strings from the regularly named functions. So that `mp_controllerAction` et. al. return nice readable output strings. Just means that we can fairly cleanly demarcate the actual code from our attempts to obfuscate it.

HelloMixpanel/MPCategoryHelpersTests/MPCategoryHelpersTests.m
samgreen: Thanks, fixed!

Mixpanel/UIView+MPHelpers.m
```diff
@@ -0,0 +1,231 @@
...
+}
+
+- (NSString *)mp_varC
+{
+    return mp_encryptHelper([self mp_imageFingerprint]);
+}
+
+- (NSArray *)mp_varSetD
+{
+    return [self mp_encryptedTargetActions];
```
alex-hofsteede: get rid of `mp_encryptedTargetActions` and just inline the code here, so we only have `mp_varSetD` and `mp_targetActions`

alex-hofsteede:
LGTM

PR #290, opened 2015-07-16
Tweaks to podspec Remove subspec from podfile since we've moved the code for the static library back in to the project itself.

samgreen:


Mixpanel.podspec
```diff
@@ -18,12 +18,6 @@ Pod::Spec.new do |s|
     ss.libraries = 'icucore', 'MPCategoryHelpers'
```
alex-hofsteede: remove `MPCategoryHelpers` here too right?

Issue #294, opened 2015-07-17
Problem of surveys for multiple users


Grubas7:
There's few problem with SDK which block Mixpanel support for surveys multiple user.

Scenario:
1. Login as UserXX
2. Show survey
3. Answer
4. Logout
5. Login as UserYY
6. Show survey
-> Survey doesn't appear

The reason of problem are:
1. `decideResponseCached` and `shownSurveyCollections` are never reseted.
2. Lack of `-flushPeople` after user select the answer.

samgreen:
Solid. Looks like an issue, thanks for the PR.

PR #295, opened 2015-07-17
Fixed problem of lack for support for surveys multiple user

Grubas7:
Fix for https://github.com/mixpanel/mixpanel-iphone/issues/294


samgreen:
Looks good to me.

PR #296, opened 2015-07-17
Cleaned out some specific provisioning profile settings and made them generic.

Stunner:
For the sample project.


samgreen:
@Stunner Thanks for the PR. Unfortunately I think this may break our continuous integration.

Stunner:
No problem.

PR #298, opened 2015-07-18
Notification/Survey Delegate Method

Stunner:
We require this in our app as we have multiple ways to display notifications and we don't want the systems colliding (i.e. we fixed an instance where Mixpanel would display a notification and our app would present another view controller immediately after, causing the notification from Mixpanel to flash). We wanted to have Mixpanel show its notification at the very beginning, but needed a way to give the app a "heads-up" so that it could halt the changing of views until the notification was dismissed. This pull request is a solution for this. Efforts were made to adhere to your coding/documentation conventions.


samgreen:
Thanks for the PRQ, we'll review and get back to you next week!

samgreen:
Hey @Stunner, we hear you loud and clear. We're going to take the spirit of this PR and make some changes with the next release. We agree that there should be an easier way to figure out what the library is going to do with respect to notifications.

I will update this issue when we've made the changes.

Stunner:
:+1:

PR #300, opened 2015-07-22
Add capability to get new notifications in real-time

ryanseams:
On behalf of a customer, submitting this PR to add support for grabbing real-time notifications. Currently the cached decide response disables users from grabbing new notifications in real-time, preventing against notifications which want to be deployed in the same session where the user becomes eligible.

Another potential solution here is to add useCache:NO into the method checkForNotificaitonsWithCompletion to force a new response each time, but have not looked into this in detail.

### DIFF
--- a/Mixpanel/Mixpanel.m
+++ b/Mixpanel/Mixpanel.m
@@ -1367,6 +1367,19 @@ static void MixpanelReachabilityCallback(SCNetworkReachabilityRef target, SCNetw
     }];
 }
 
+- (void)checkForNewNotificaitonsWithCompletion:(void (^)(BOOL available))completion
+{ 
+    [self checkForDecideResponseWithCompletion:^(NSArray *surveys, NSArray *notifications, NSSet *variants, NSSet *eventBindings) {
+        if (completion) {
+            BOOL available = NO;
+            if(notifications.count){
+                available = YES;
+            }
+            completion(available);
+        }
+    } useCache:NO];
+}
+
 - (void)checkForVariantsWithCompletion:(void (^)(NSSet *variants))completion
 {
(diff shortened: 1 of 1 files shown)


Mixpanel/Mixpanel.m
```diff
@@ -1367,6 +1367,19 @@ - (void)checkForNotificationsWithCompletion:(void (^)(NSArray *notifications))co
     }];
 }
 
+- (void)checkForNewNotificaitonsWithCompletion:(void (^)(BOOL available))completion
+{
```
samgreen: This looks like the same code as `checkForNotificationsWithCompletion:` only just returning a `BOOL`. Couldn't the customer just implement this in their code? Additionally, I'm concerned that the name of the function might be confusing for developers.

diggoryR:
Also, I'm sure this was noted, but would definitely want to fix the "Notificaitons" typo on line 1370

samgreen:
Good catch @diggoryR

samgreen:
Also this change isn't exposed in `Mixpanel.h` so there is no way for external developers to access it.

ryanseams:
Really this was simply a copy paste of his main code.

The bigger issue here is people want a way to grab notifications in real-time -- the current problem is the cached response prevents against this when running checkForDecideResponseWithCompletion.

The other method I have seen is to invalidate the cache in the identify method to force new notifs to show -- but this has battery drain concerns.

Is there a best practice for adding this? Maybe we make a config option which then allows for real-time versus passive grabbing of notifications? Happy to propose some other solutions as needed.

samgreen:
We're going to look at altering the way we fetch notifications. I'm working on something now and I will update this issue when we have a PR nearing completion.

ryanseams:
Thanks @samgreen appreciate you taking this one head on!

PR #304, opened 2015-07-28
Flip app release and version to be correct

ryanseams:
These are backwards and causing issues when segmenting by these properties to view cross-platform statistics.

You may want to review this from your end as far as impact to current projects -- this will change the props going forward to be correct, but will not fix those which are currently using the wrong behavior.

### DIFF
--- a/Mixpanel/Mixpanel.m
+++ b/Mixpanel/Mixpanel.m
@@ -942,8 +942,8 @@ static __unused NSString *MPURLEncode(NSString *s)
     CTCarrier *carrier = [self.telephonyInfo subscriberCellularProvider];
 
     // Use setValue semantics to avoid adding keys where value can be nil.
-    [p setValue:[[NSBundle mainBundle] infoDictionary][@"CFBundleVersion"] forKey:@"$app_version"];
-    [p setValue:[[NSBundle mainBundle] infoDictionary][@"CFBundleShortVersionString"] forKey:@"$app_release"];
+    [p setValue:[[NSBundle mainBundle] infoDictionary][@"CFBundleVersion"] forKey:@"$app_release"];
+    [p setValue:[[NSBundle mainBundle] infoDictionary][@"CFBundleShortVersionString"] forKey:@"$app_version"];
     [p setValue:[self IFA] forKey:@"$ios_ifa"];
     [p setValue:carrier.carrierName forKey:@"$carrier"];
     [p setValue:[self watchModel] forKey:@"$watch_model"];
--- a/Mixpanel/Mixpanel.m
+++ b/Mixpanel/Mixpanel.m
@@ -942,8 +942,10 @@ static __unused NSString *MPURLEncode(NSString *s)
     CTCarrier *carrier = [self.telephonyInfo subscriberCellularProvider];
 
     // Use setValue semantics to avoid adding keys where value can be nil.
-    [p setValue:[[NSBundle mainBundle] infoDictionary][@"CFBundleVersion"] forKey:@"$app_release"];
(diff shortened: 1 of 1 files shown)


samgreen:
@ryanseams Hmm, this will drastically change data for existing clients. Even if these are switched, I'm not sure it's going to be okay to swap them back.

ryanseams:
Is the fix instead to change the mixpanel-android lib to reflect this behavior?

Currently cross-platform apps cannot segment by these properties as they are not the same in each SDK -- they are mirror images of one another

alex-hofsteede:
Based on a reading of [this](https://developer.apple.com/library/ios/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102364) and the descriptions of the variables, it seems like 

app_release == CFBundleShortVersionString
app_version == CFBundleVersion 

Is the more correct way.

samgreen:
@alex-hofsteede I'm more than happy to merge this chance, but concerned about the impact for current customers.

alex-hofsteede:
@samgreen no I'm saying the way it was before seems like the correct way.

samgreen:
Ah I see, so then we just have a mismatch between the iOS and Android libs.

ryanseams:
@alex-hofsteede should i make this instead a pull on Android to make this same change?

I already have one IP for adding the missing of these two: https://github.com/mixpanel/mixpanel-android/pull/244

alex-hofsteede:
hold on a sec, maybe I was wrong. I misread the PR

http://cl.ly/image/2k452Q2W2e1y in that screenshot. the "Version" becomes CFBundleShortVersionString, which we should be assigning to app_version. So yeah, I guess this PR is correct, but we still need to figure out what the impact would be on existing integrations before we can make this change,

alex-hofsteede:
actually the properties should maybe be something like app_version and app_build. but yeah, can we make this change without breaking a ton of peoples funnels.

samgreen:
The only way I can see to not break existing funnels, is to create 2 new properties where the names match for both Android and iOS. Not great.

alex-hofsteede:
Yeah, maybe the only way to do this without breaking things is to come up
with 2 new properties and add them to both libs in parallel with the ones
that are already there. Then we'd have to wait like a year or so before
retiring the old ones.

alex-hofsteede:
Sadly this mistake goes all the way back to 85df778d when we added the build number (which is admittedly badly named as CFBundleVersion) as the app version

ryanseams:
Should we do something like adding $app_release_num and $app_version_num or stray away from the same naming convention and add entirely newly named properties here? Happy to make the PR for both libs if we can agree on a convention

alex-hofsteede:
@ryanseams yeah I think 2 new properties is the only real way, we can't break old integrations. I would go for something like `app_version_string` and `app_build_number`, let's get away from 'release' as it's confusing and we basically have no idea whether a given version is actually a 'release' from the customers perspective anyway. All depends on what the terminology for android is too. 

Also the other reason we have to keep the old properties is because of our trends reports, which would break if we changed these.

samgreen:
@ryanseams That would be super helpful. We just need to decide on names. I'm open to the ones you suggested. @alex-hofsteede?

ryanseams:
I updated the PR to add two new props, ```app_version_string``` and ```app_build_number``` as suggested by @alex-hofsteede.

The only issue I forsee is with the UI -- how will the current "App Version" differ from this "App Version" in our display? Should we instead go with ```app_build_string``` or something which has a different name so it is easy to tell the difference within the UI?

Issue #303, opened 2015-07-26
Events never flush after a crash (2.8.2)


idStar:
I've noticed consistently while building/testing my crash reporting, that event activity is no longer sent to Mixpanel once a crash has occurred.

##### How to Reproduce
1. Ensure normal activity in your app is sending a steady stream of event activity to Mixpanel.
2. Raise an exception in your code, purposely, that you don't catch. In my case, I was using [FCYAsserts](https://github.com/fcy/FCYAsserts), but you could use [Marcus Zarra's ZAssert()](http://www.cimgf.com/2010/05/02/my-current-prefix-pch-file/) or other similar, assertion macros that raise exceptions.
3. Without rebuilding your app, re-launch it and choose a different pathway through it that won't crash, but which should generate additional tracking events. 

##### Discussion

It's when doing Step 3, even hours after the crash and subsequent app activity, that no matter what I did, no events would come in from the app.

I would have to lay in a fresh copy of the app by building from Xcode onto the device, to get the Mixpanel queue to be 'reset'. However, previously _jammed_ events were lost. I would never see them show up on the Mixpanel dashboard.

I suspect that after a crash, the Mixpanel archive of events is in some corrupted, unrecoverable state.

I have enough logic in my app that I know if a launch was preceded by a crash. At such junctures, is there some way I can force a reset of the Mixpanel cache, so I can get back to a reliable state?

samgreen:
This is a duplicate of #251

samgreen:
This should be resolved with #306

patedit:
Solution proposed on #306 works with current Crashlytics library and both exception handlers can co-exist at the same time. It will be included in the next release v2.9.2. Meanwhile, feel free to checkout #306 and use it.
We're not able to reproduce the lost of events after a crash following your steps. Could you give some more information?

idStar:
@patedit I'm now putting together a sample app to recreate this phenomenon. Sorry for the delay. As mentioned earlier, I'm starting to think this is going to turn out to be the combination of {HockeyApp, Mixpanel} whereas {Crashlytics, Mixpanel} seem to now place nice vis-a-vis top level uncaught exception handlers. Will post findings within a day or so (have to setup profiles for each service for my demo project to illustrate this).

idStar:
##### Update
* Creating a fresh test app (attached) with Cocoapods, and using the latest official HockeySDK-source pod and Mixpanel cocoapod worked well. 
[...]

PR #306, opened 2015-07-29
Added MixpanelExceptionHandler.

samgreen:


alex-hofsteede:
Can we create a unit test for this functionality? or is there no real way to assert something after forcing a crash?

idStar:
@samgreen Not sure if more is coming in this branch as I see your pull request. I just tested this branch and events are still jammed after a crash. I'll keep testing as you have updates.

events_are_still_jammed

idStar:
@samgreen I've noticed the SDK when from version `2.8.3` to `2.9.0`, but there's some serious issues in Milestone 2.8.4 (notably this one) that have had no movement in months. What's the plan? 

This is a serious bug and limitation of the SDK as it stands right now.

idStar:
It looks like Milestone 2.9.1 has been released (https://github.com/mixpanel/mixpanel-iphone/releases/tag/v2.9.1) but this issue didn't make it in. It's been open since July 2015 and is quite serious. Can someone advise on what the realistic roadmap for this is?

samgreen:
Hey @idStar so sorry for the churn on this. We've been super distracted here, but we're taking another look and hoped to get this resolved in Jan.

idStar:
@samgreen Per my comment in Issue #303 [here](https://github.com/mixpanel/mixpanel-iphone/issues/303#issuecomment-167680898), would it be possible to rebase this branch from master, so that modern Carthage support is available to it? 

I want to try this branch in my own (real) app, but it's not coming in cleanly with Carthage, as I believe this branch predates some more robust Carthage support. Thanks.

idStar:
Following up for documentation purposes, the problem I posted above is no longer an issue for me, as tested with the current code in `master` of this repo, and me having switched to eagerly initializing my `Mixpanel` sharedInstance, instead of lazily (not sure why that would matter, but it helped).

Full details in ticket #303: https://github.com/mixpanel/mixpanel-iphone/issues/303#issuecomment-168109441

samgreen:
Merging this for next release.

PR #307, opened 2015-07-29
Revert the NS_DESIGNATED_INITIALIZER macro for now.

### LINKED ISSUE #305: New Warnings Under Xcode 7 - Designated Initializer
In many places, it looks like Mixpanel adds an initializer but doesn't override the designated initializer. This is causing new warnings under Xcode 7 all over the place.

Example of warning:
/Users/ccornaby/discover-ios9/SDKs/Mixpanel/Mixpanel/MPPropertyDescription.m:54:17: Method override for the designated initializer of the superclass '-init' not found
/Applications/Xcode-beta.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS9.0.sdk/usr/include/objc/NSObject.h:60:1: Method marked as designated initializer of the class here

I'm adding Mixpanel to my project as raw source, we don't use CocoaPods.

samgreen:
Resolves #305


Coeur:
I would prefer to override the superclass designated initializers. I will attempt a patch.

PR #308, opened 2015-07-29
Update README for Carthage

samgreen:


README.md
```diff
@@ -1,16 +1,45 @@
...
 
 The Mixpanel library for iOS is an open source project, and we'd love to see your contributions! We'd also love for you to come and work with us! Check out http://boards.greenhouse.io/mixpanel/jobs/25226#.U_4JXEhORKU for details.
 
-# Painless Installation
+# Painless Installation (CocoaPods)
 
 Mixpanel supports `Cocoapods` for easy installation.
 To Install, see our **[full documentation »](https://mixpanel.com/help/reference/ios)**
 
+# Carthage
```
alex-hofsteede: So this is pretty in depth, whereas the cocoapods one is just like "we do cocoapods!, click here for details" 

Maybe we should have inline cocoapods instructions here too, so they are all in one place. But the tradeoff of adding all this stuff inline is that the README starts looking like you need a CS degree to add Mixpanel to your app. In the shoes of an iOS app dev, what are your thoughts? What are some iOS projects with particularly good READMEs that you like?

samgreen: You know that's a great point. I ripped the docs wholesale from Carthage anyway, so I'll just link to it instead.

PR #309, opened 2015-07-31
Default the flush timer to 1 second instead of 60 seconds for debug mode.

samgreen:


Mixpanel/Mixpanel.m
```diff
@@ -106,7 +106,14 @@ + (Mixpanel *)sharedInstanceWithToken:(NSString *)apiToken launchOptions:(NSDict
 {
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
-        sharedInstance = [[super alloc] initWithToken:apiToken launchOptions:launchOptions andFlushInterval:60];
+        
+#if defined(DEBUG)
```
jbwyme: should this code be against the left margin?

samgreen: Yes, as a matter of style in Mixpanel's Objective-C code compiler directives like `#ifdef` are never indented.

jbwyme: weird ok

jbwyme:
lgtm

Suhail:
thank you @samgreen - only real issue during iOS integration that really confused me.

Grubas7:
IMHO this might be confusing for developers/testers.
Most of the time my team is testing app in debug mode, so thanks this change SDK/app will act different in debug and release mode.
I think if someone want such feature should do this in app not in SDK.

Beside flush will be never executed, because flush timer is not fired, so someone will have to set `-flushInterval` anyway.

samgreen:
@Grubas7 This change is to make it less confusing for new integrations. Previously they would have to wait a full minute if they're watching the live view before seeing any activity. 

If you're concerned this will affect you negatively, you can use the alternative constructor and set your own flush interval. The flush timer is started automatically in `applicationDidBecomeActive:`.

This behavior also now matches our Android library, and feature parity is always a good thing.

PR #310, opened 2015-08-04
Add documentation for Cocoapods

Palleas:
In case the user of the SDK wants to activate the logging and debugging modes without tweaking the Pod directly and Xcode and loosing his changes for every `pod install`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,28 @@ The Mixpanel library for iOS is an open source project, and we'd love to see you
 Mixpanel supports `Cocoapods` for easy installation.
 To Install, see our **[full documentation »](https://mixpanel.com/help/reference/ios)**
 
+If you choose to go with this method and want to activate the logging and debugging modes in the SDK
+(see "Debugging and logging" in **[the documentation »](https://mixpanel.com/help/reference/ios)**), you can use a
+`post_install` hook in your Podfile, like so:
+
+```ruby
+post_install do |installer|
+  installer.pods_project.targets.each do |target|
+    target.build_configurations.each do |config|
+      settings = config.build_settings['GCC_PREPROCESSOR_DEFINITIONS']
+      settings = ['$(inherited)'] if settings.nil?
+
+      if target.name == 'Pods-MyProject-Mixpanel'
+        settings << 'MIXPANEL_DEBUG=1'
+        settings << 'MIXPANEL_ERROR=1'
+      end
+
+      config.build_settings['GCC_PREPROCESSOR_DEFINITIONS'] = settings
+    end
+  end
+end
+```
+
 # Manual Installation
 
 To help users stay up to date with the latests version of our iOS SDK, we always recommend integrating our SDK via CocoaPods, which simplifies version updates and dependency management. However, there are cases where users can't use CocoaPods. Not to worry, just follow these manual installation steps and you'll be all set.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,6 +11,28 @@ The Mixpanel library for iOS is an open source project, and we'd love to see you
...
+    installer.pods_project.targets.each do |target|
+        target.build_configurations.each do |config|
+            s = config.build_settings['GCC_PREPROCESSOR_DEFINITIONS']
+            if s==nil then s = [ '$(inherited)' ] end
+            
+            if target.name == "Pods-MyProject-Mixpanel"
+                s.push('MIXPANEL_DEBUG=1')
+                s.push('MIXPANEL_ERROR=1')
+            end
+            s.push(second_option)
```
remiprev: What is `second_option` here?

Palleas: That would be a fail in my copy/paste process :smile:

Palleas: Fixed!

README.md
```diff
@@ -11,6 +11,27 @@ The Mixpanel library for iOS is an open source project, and we'd love to see you
...
+            if s==nil then s = [ '$(inherited)' ] end
+
+            if target.name == "Pods-MyProject-Mixpanel"
+                s.push('MIXPANEL_DEBUG=1')
+                s.push('MIXPANEL_ERROR=1')
+            end
+            config.build_settings['GCC_PREPROCESSOR_DEFINITIONS'] = s
+        end
+    end
+end
```
remiprev: I took the liberty of rewriting your Ruby code to something more readable and similar to what the Ruby community agrees on :smile: 

```ruby
post_install do |installer|
  installer.pods_project.targets.each do |target|
    target.build_configurations.each do |config|
      settings = config.build_settings['GCC_PREPROCESSOR_DEFINITIONS']
      settings = ['$(inherited)'] if settings.nil?

      if target.name == 'Pods-MyProject-Mixpanel'
        settings << 'MIXPANEL_DEBUG=1'
        settings << 'MIXPANEL_ERROR=1'
      end

      config.build_settings['GCC_PREPROCESSOR_DEFINITIONS'] = settings
    end
  end
end
```

README.md
```diff
@@ -11,6 +11,28 @@ The Mixpanel library for iOS is an open source project, and we'd love to see you
...
+If you choose to go with this method and want to activate the logging and debugging modes in the SDK
+(see "Debugging and logging" in **[the documentation »](https://mixpanel.com/help/reference/ios)**), you can use a
+`post_install` hook in your Podfile, like so:
+
+```ruby
+post_install do |installer|
+    installer.pods_project.targets.each do |target|
+        target.build_configurations.each do |config|
+            s = config.build_settings['GCC_PREPROCESSOR_DEFINITIONS']
+            if s==nil then s = [ '$(inherited)' ] end
```
kylef: You can shorthand this to:

```ruby
s = [ '$(inherited)' ] unless s
```

Slightly more readable and simpler.

README.md
kylef: `second_option` doesn't seem to be defined anywhere?

README.md
Palleas: I trust you on this :stuck_out_tongue: 
What about @kylef's `s ||= [ '$(inherited)' ]`, that looks nice!

samgreen:
Hey thanks for the PR! I like this change and it should be documented, but our README is getting very long. I would prefer to put this bit of ruby code on the docs at Mixpanel.com. What do you think?

Palleas:
I don't see why not, it would fit perfectly under the "Debugging and logging" section :+1:

samgreen:
We've updated the iOS integration docs now. Thanks!

Palleas:
Nice! I can't see it in the doc yet though, I'm guessing you have to deploy or something like that?

samgreen:
@Palleas Should be live by the end of the day!

PR #314, opened 2015-08-16
renaming Messsage to Message

Coeur:
fixing a spelling mistake, deprecating old name

### DIFF
--- a/Mixpanel/MPDesignerEventBindingRequestMessage.m
+++ b/Mixpanel/MPDesignerEventBindingRequestMessage.m
@@ -1,5 +1,5 @@
 //
-//  MPDesignerEventBindingRequestMesssage.m
+//  MPDesignerEventBindingRequestMessage.m
 //  HelloMixpanel
 //
 //  Created by Amanda Canyon on 7/15/14.
@@ -55,7 +55,7 @@ NSString *const MPDesignerEventBindingRequestMessageType = @"event_binding_reque
 
 @end
 
-@implementation MPDesignerEventBindingRequestMesssage
+@implementation MPDesignerEventBindingRequestMessage
 
 + (instancetype)message
 {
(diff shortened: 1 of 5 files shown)


Mixpanel/MPDesignerEventBindingResponseMessage.m
```diff
@@ -26,3 +26,7 @@ - (NSString *)status
 }
 
 @end
+
+@implementation MPDesignerEventBindingResponseMesssage
```
samgreen: Looks like you missed one

Mixpanel/MPDesignerEventBindingRequestMessage.m
```diff
@@ -88,3 +88,7 @@ - (NSOperation *)responseCommandWithConnection:(MPABTestDesignerConnection *)con
 }
 
 @end
+
+@implementation MPDesignerEventBindingRequestMesssage
```
samgreen: This one is still wrong

samgreen:
I do not believe you can mark a class as `__deprecated`, @alex-hofsteede can you chime in on this?

samgreen:
I'm also unable to find any reference to marking a class with the deprecated attribute from the LLVM manual. http://clang.llvm.org/docs/AttributeReference.html

Coeur:
@samgreen http://stackoverflow.com/questions/7408980/how-can-i-deprecate-an-entire-protocol and http://stackoverflow.com/questions/3908715/how-do-i-flag-a-method-as-deprecated-in-objective-c-2-0/12407525#12407525

samgreen:
Just tried this in Xcode 6 and I don't get a warning using the deprecated class. Any idea why?

Coeur:
Maybe not supported by Xcode. :/

PR #315, opened 2015-08-16
renaming bindng to binding

Coeur:
fixing spelling mistake and marking old method as deprecated


samgreen:
Good catch, thanks!

Coeur:
It's similar to https://github.com/mixpanel/mixpanel-iphone/pull/314

PR #317, opened 2015-08-16
fixing Xcode 7 warnings related to NS_DESIGNATED_INITIALIZER

Coeur:
This patch is following my own recommandation on the matter: https://stackoverflow.com/questions/29639040/warning-method-override-for-designated-initializer/31830471#31830471

If approved, it supersedes https://github.com/mixpanel/mixpanel-iphone/pull/307


Coeur:
As #307 got merged instead, this current pull request can be closed as it does the opposite of #307.

PR #325, opened 2015-08-27
Reset `shownNotifications` on Mixpanel#reset call

yas375:
In continuation of https://github.com/mixpanel/mixpanel-iphone/issues/232#issuecomment-135206887


samgreen:
Solid. Thanks!

samgreen:
Solid. Thanks!

PR #326, opened 2015-08-28
Add deploy.sh to help automatically deploy new releases

samgreen:


deploy.sh
```diff
@@ -0,0 +1,47 @@
...
+
+rm -rf ./Docs/HeaderDoc
+headerdoc2html -q -o ./Docs ./Mixpanel/Mixpanel.h > /dev/null 2>&1
+mv ./Docs/Mixpanel_h ./Docs/HeaderDoc
+cp -r ./Docs/HeaderDoc/* ../analytics/media/doctyl/uploads/iPhone-spec/
+
+echo "Updated docs in ../analytics. Ensure you commit and PR from that repo as well!"
+
+echo "Creating new tag vX.X.X"
+git tag "v$INCREMENTED_VERSION"
```
zakj: don't you need to commit the changes to `Mixpanel/Mixpanel.m` and `Mixpanel.podspec` before tagging?

deploy.sh
```diff
@@ -0,0 +1,47 @@
...
+git push origin "v$INCREMENTED_VERSION"
+
+echo "Verifying podspec"
+pod spec lint
```
zakj: maybe lint before commit/tagging?

deploy.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+echo "Bumping version number."
+VERSION=$(grep -o "VERSION @\"[0-9]*\.[0-9]*\.[0-9]*" Mixpanel/Mixpanel.m | tail -1 | tr -d 'VERSION @"')
+VERSION_MAJOR=$(echo $VERSION | awk -F \. '{print $1}')
+VERSION_MINOR=$(echo $VERSION | awk -F \. '{print $2}')
+VERSION_PATCH=$(echo $VERSION | awk -F \. '{print $3}')
+
+if [ $VERSION_PATCH ==  9 ]; then
+    if [ $VERSION_MINOR ==  9 ]; then
```
evnp: We never go higher than 9 for minor versions?

alex-hofsteede: yeah that's not right. I mean, we haven't gone above 9 before, but we almost certainly will for 2.x.x

samgreen: Should I ditch the automation and let people enter a new version? Something like:

`Current Version: 2.8.3`
`New Version (X.X.X): `

zakj: Similar tools take `--major`/`--minor`/`--patch` style command-line arguments to determine how to bump.

PR #330, opened 2015-08-31
Fixes for Xcode 7 warnings

colincornaby:
I'm getting a few warnings under Xcode 7 (we're not Cocoapods, so I'm using a submodule.) These are some fixes, it's all just compiler hints.


colincornaby:
The build died but I think that's an issue build side. This should all still work under earlier versions of Xcode, but I'll check.

danieljimenez:
+1, this fixed it for us.

samgreen:
Merged another PR that checks compiler version.

PR #337, opened 2015-09-04
Sets deploy target to 8 across the board for framework target.

schwa:
iTunes Connect will reject any app linking to dyn framework of Mixpanel.

Framework builds do not support older than iOS 8


samgreen:
Thanks!

PR #340, opened 2015-09-12
Add property to customize in-app mini notifications background

mtucker:
* Added a new property called miniNotiicationBackgroundColor to primary interface
* Setting this property will set the background color of the notification view
* If this property is not set, we fall back to previous behavior

### DIFF
--- a/Mixpanel/Mixpanel.h
+++ b/Mixpanel/Mixpanel.h
@@ -201,6 +201,19 @@
 
 /*!
  @property
+ 
+ @abstract
+ If set, determines the background color of mini notifications.
+
+ @discussion
+ If this isn't set, we default to either the color of the UINavigationBar of the top 
+ UINavigationController that is showing when the notification is presented, the 
+ UINavigationBar default color for the app or the UITabBar default color.
+ */
+@property (atomic) UIColor* miniNotificationBackgroundColor;
+
+/*!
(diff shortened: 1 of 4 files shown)


Mixpanel/Mixpanel.h
```diff
@@ -201,6 +201,19 @@
...
+ @abstract
+ If set, determines the background color of mini notifications.
+
+ @discussion
+ If this isn't set, we default to either the color of the UINavigationBar of the top 
+ UINavigationController that is showing when the notification is presented, the 
+ UINavigationBar default color for the app or the UITabBar default color.
+ */
+@property (atomic) UIColor* miniNotificationBackgroundColor;
+
```
samgreen: Should be `strong`

mtucker: Sorry, I thought `strong` was the default?

From the [docs](https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/EncapsulatingData/EncapsulatingData.html#//apple_ref/doc/uid/TP40011210-CH5-SW30):

>There’s no need to specify the strong attribute explicitly, because it is the default.

samgreen: We specify explicitly.

PR #342, opened 2015-09-14
Fix for xcode 7

rpranata:
xcode throws a warning but we have "Treat warnings as errors" options on. So, it wont actually compile

### DIFF
--- a/Mixpanel/MPNotificationViewController.m
+++ b/Mixpanel/MPNotificationViewController.m
@@ -147,7 +147,11 @@
     return UIStatusBarAnimationFade;
 }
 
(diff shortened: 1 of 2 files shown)


rpranata:
The build check failed for this command. I dont have any idea why this is failing :(

`The command "openssl aes-256-cbc -k "$ENCRYPTION_SECRET" -in scripts/profile/HelloMixpanel_Ad_Hoc.mobileprovision.enc -d -a -out scripts/profile/HelloMixpanel_Ad_Hoc.mobileprovision" failed and exited with 1 during .`

samgreen:
The error is not your fault. It's because you're not in the org. Thanks for the PR!

PR #344, opened 2015-09-17
Add delegate for DidSetIdentify

elihayc:
The JoinExperiement has Async call to SetIdentify.
And If we want to use experiment on the startup, we might check the tweak before the setIdentify Completed.

I added a delegate  "mixpanelDidSetIdentify" , which called when the SetIdentify method completed.


Mixpanel/Mixpanel.h
```diff
@@ -948,4 +948,18 @@
...
 
+/*!
+ @method
+ 
+ @abstract
+ Raise after the identify sets asynchronously.
+ 
+ @discussion
+ Execute your code after the identify set.
+ It is usefull for getting the ABTesting MPTweakValue (and keep it consistent value for each user identify)
```
samgreen: Typo, should be "useful"

Mixpanel/Mixpanel.m
```diff
@@ -359,6 +359,17 @@ - (void)identify:(NSString *)distinctId
         if ([Mixpanel inBackground]) {
             [self archiveProperties];
         }
+        
+        // Clear cached variants
+        self.variants = nil;
```
samgreen: Why are you clearing the variants here?

Mixpanel/Mixpanel.h
elihayc: fixed

Mixpanel/Mixpanel.m
elihayc: Because if i'm not, the variants cached data of the previous user applied on the current user.

I also have a sample app for simulate this issue, if you want, I can send it to you.
BTW, I worked with Orien Wu from Mixpanel to solve this issue

samgreen: @elihayc  You need to call `stop` on all of the currently executing variants. You're just removing references to them here, and they will still be applied. If you can make this change, I'm happy to merge it in.

elihayc: @samgreen , I changed it to stop instead of removing the reference.

samgreen:
Upon giving this a second look, I don't think we want to call `joinExperimentsWithCallback:` inside of `identify:` as this has a huge potential to break with A/B tests that are using an alias.

elihayc:
@samgreen,  Do you have another idea / solution for getting the user experiments right after we call to SetIdentify?
Looks to me like a mandatory feature for signout-> sign in (with another user) flow.

PR #363, opened 2015-10-19
Changed all usage of MixpanelDebug macro to MessagingDebug in MPWebSocket.m

jaredmixpanel:
Currently MPWebSocket.m continuously spams the console with debug logging enabled. These messages should use the MessagingDebug macro instead of MixpanelDebug.


samgreen:
Thanks

Issue #346, opened 2015-09-22
Integrating Mixpanel as a Framework breaks Tweaks


arittr:
Integrating Mixpanel with CocoaPods v0.36+ and use_frameworks prevents Tweaks from appearing in the UI (or being found / stored in general). 

It appears the the issue has something to do with the macro not being included in the output framework - the relevant line is https://github.com/mixpanel/mixpanel-iphone/blob/master/Mixpanel/MPTweakInline.m#L127 , which will return null when using frameworks. It looks like the whole function is just undefined in the final SDK binary:

"@MPTweak *_MPTweakCreateWithEntry(NSString *__strong, mp_tweak_entry *)^@<Unknown Function>

RTimal:
I'm experiencing this as well

samgreen:
#364 should fix that

phin:
+1, experiencing this issue as well

samgreen:
This is now released as v2.9.2, thanks for your patience!

z4r:
I still have the same issue with the v2.9.3 and v2.9.4 installed via Carthage.
The visibility is provided only for the main header `Mixpanel.h` and any symbol related to `MPTweakInline.h` is not available.

![Mixpanel Header](https://cloud.githubusercontent.com/assets/639459/13325925/336e97f6-dbe4-11e5-9bd1-7426f7b8cad5.png)

PR #364, opened 2015-10-20
Update MPTweakInline to handle multiple dynamically loaded images.

samgreen:


projectAZB:
Hey Sam, my team and I were experiencing the exact problem detailed in #346, found your solution, replaced the contents of Mixpanel/MPTweakInline.m with your fix, and everything worked perfectly. I was wondering why this hasn't been merged into the SDK yet, and what sort of errors you could foresee going forward with your unmerged fix integrated into the project.

samgreen:
Hi @projectAZB, glad that the fix in #364 works for you. We originally wrote the fix as a means to allow tweaks in Swift. Unfortunately adding tweaks to Swift ended up being a lot more tricky than we initially estimated. That's why the original PR was never merged and released.

I can see how the PR would also resolve the original issue and should be merged in for the next release ASAP. I don't foresee any issues in using this code moving forward. I believe it to be more hardened and more reliable as it will iterate over all of the loaded images in memory. I've tagged that PR for the next release and I'll update you when it is merged.

projectAZB:
Perfect, this was a huge help, thanks for your response!

rdbayer:
I'm using build 2.9.2, but I'm still getting this error when I try to use MPTweakValue():

```'MPTweakValue' (aka 'AnyObject') cannot be constructed because it has no accessible initializers.```

@samgreen Is MPTweakValue("x", 5) meant to work now in Swift, or is there a workaround I'm meant to use?

PR #367, opened 2015-10-20
Open url after MPTakeoverNotificationViewController is dismissed

yas375:
We are using takeover (full screen) notifications. When user interacts with them Mixpanel sdk launches an url which is handled by our app. 

In our case those urls are handled by presenting some view controller on top of current top presented controller.

The problem we hit pretty often is that the our view controller is not presented. And in console I see something like:

> Warning: Attempt to present <UINavigationController: 0x134a5b600> on <MPTakeoverNotificationViewController: 0x136a37650> whose view is not in the window hierarchy!

I suppose this happens because in current implementation `[[UIApplication sharedApplication] openURL:url]` is called first and after that `MPTakeoverNotificationViewController` get dismissed. So at the time our url handled gets called `MPTakeoverNotificationViewController` is not completely dismissed and is still top presented view controller.

With this fix the problem seems to be solved :)

### DIFF
--- a/Mixpanel/Mixpanel.m
+++ b/Mixpanel/Mixpanel.m
@@ -1758,16 +1758,17 @@ static void MixpanelReachabilityCallback(SCNetworkReachabilityRef target, SCNetw
     };
 
     if (status && controller.notification.callToActionURL) {
-        MixpanelDebug(@"%@ opening URL %@", self, controller.notification.callToActionURL);
-        BOOL success = [[UIApplication sharedApplication] openURL:controller.notification.callToActionURL];
+        [controller hideWithAnimation:YES completion:^{
+            NSURL *URL = controller.notification.callToActionURL;
+            MixpanelDebug(@"%@ opening URL %@", self, URL);
 
-        [controller hideWithAnimation:!success completion:completionBlock];
+            if (![[UIApplication sharedApplication] openURL:URL]) {
+                MixpanelError(@"Mixpanel failed to open given URL: %@", URL);
+            }
 
-        if (!success) {
-            MixpanelError(@"Mixpanel failed to open given URL: %@", controller.notification.callToActionURL);
-        }
-
-        [self trackNotification:controller.notification event:@"$campaign_open"];
+            [self trackNotification:controller.notification event:@"$campaign_open"];
+            completionBlock();
+        }];
     } else {
         [controller hideWithAnimation:YES completion:completionBlock];
     }


Mixpanel/Mixpanel.m
```diff
@@ -1667,16 +1667,17 @@ - (void)notificationController:(MPNotificationViewController *)controller wasDis
     };
 
     if (status && controller.notification.callToActionURL) {
-        MixpanelDebug(@"%@ opening URL %@", self, controller.notification.callToActionURL);
-        BOOL success = [[UIApplication sharedApplication] openURL:controller.notification.callToActionURL];
+        [controller hideWithAnimation:NO completion:^{
+          NSURL *url = controller.notification.callToActionURL;
```
samgreen: Please rename this to `URL`

samgreen:
Please fix the indentation with `Control + I`. We use 4 spaces.

Mixpanel/Mixpanel.m
yas375: updated

yas375:
Oops, my bad with the indentation. Fixed :)

Mixpanel/Mixpanel.m
```diff
@@ -1667,16 +1667,17 @@ - (void)notificationController:(MPNotificationViewController *)controller wasDis
     };
 
     if (status && controller.notification.callToActionURL) {
-        MixpanelDebug(@"%@ opening URL %@", self, controller.notification.callToActionURL);
-        BOOL success = [[UIApplication sharedApplication] openURL:controller.notification.callToActionURL];
+        [controller hideWithAnimation:NO completion:^{
```
yas375: I'd like to bring some attention to `NO` here. I don't have a strict reason why to dismiss without animation...

The previous logic was: if `UIApplication#openURL:` returns `YES` - then dismiss without animation.

With this PR we are opening the url after it got dismissed. I assume that in most cases of using this library `openURL:` will return `YES` because people are going to use this notifications to show something in the same app or to open a Safari or make a phone call... I don't use `UIApplication#canOpenURL:` because of the notes in the documentation about iOS9. Assuming this library is used by a lot of developers in lots of different apps I think we shouldn't rely on `canOpenURL:`.

If the url is `http` or `tel`, then it's probably fine to dismiss the notification panel without animation and then the app goes into background. If the url is within the same app, then it might make sense to dismiss with animation. As I said before I don't have a strict opinion on this. But maybe someone else has ;)

yas375:
Any feedback on this? I'll be on a vacation next week so I won't be able to apply any feedback. Still can do something today. Thanks in advance.

yas375:
Any news on this? The last response was more than 2 month ago and I've applied all the feedback.

Mixpanel/Mixpanel.m
samgreen: My only concern with this is that if a developer adds a custom URL to their notification they may not be aware that we are using this method internally and if they haven't registered the custom URL in the Info.plist then this code change will cause it to silently fail.

So the question now, is should we change this behavior using the new method and document what's happening so developers will understand, or should we use the previous `openURL:` call which seems to have the same behavior on iOS 9.

samgreen:
So sorry @yas375 We've been heads down on other things, I took a second look now and left some feedback. I would love to get this resolved and merged this week.

Mixpanel/Mixpanel.m
yas375: I apology for the delayed response.

> My only concern with this is that if a developer adds a custom URL to their notification they may not be aware that we are using this method internally and if they haven't registered the custom URL in the Info.plist then this code change will cause it to silently fail.

This would be the case if we start using `canOpenURL:` before opening the URL. But we don't do that here. We still use same `openURL:` which seem to have the same behavior.

I think it's correct here to use `openURL:` as before. The only thing I didn't have a strong opinion on is whether to dismiss notification view controller with animation or without. We have shipped this code in our app a few month ago and didn't have any problems with it. I think it's ok to merge without animation and if somebody will bring a question, it can be changed then.

samgreen: Yes you are totally correct. Sorry for the confusion about the iOS 9 URL changes. I would very much prefer to dismiss the controller with an animation. From a UX perspective I think it's rather jarring to dismiss with no animation. 

If we could dismiss like before based on the success, I would be satisfied merging this today.

yas375: Agree. Have changed to dismiss with animation. Also rebased on master and force pushed.

samgreen:
Thanks!

PR #370, opened 2015-10-23
Retry flushing events in a greater number of failure cases, such as w…

### LINKED ISSUE #369: Retry flushing after errors functioning correctly?
Hi –

First of all, thanks for making this an Open Source library.

I just looked briefly at the code trying to figure out what the retry behavior is when tracking events in case the user is on a bad network.

The relevant code is this:

```
- (void)flushQueue:(NSMutableArray *)queue endpoint:(NSString *)endpoint
{
    while ([queue count] > 0) {
        NSUInteger batchSize = ([queue count] > 50) ? 50 : [queue count];
        NSArray *batch = [queue subarrayWithRange:NSMakeRange(0, batchSize)];

        NSString *requestData = [self encodeAPIData:batch];
        NSString *postBody = [NSString stringWithFormat:@"ip=1&data=%@", requestData];
        MixpanelDebug(@"%@ flushing %lu of %lu to %@: %@", self, (unsigned long)[batch count], (unsigned long)[queue count], endpoint, queue);
        NSURLRequest *request = [self apiRequestWithEndpoint:endpoint andBody:postBody];
        NSError *error = nil;

        [self updateNetworkActivityIndicator:YES];

        NSURLResponse *urlResponse = nil;
        NSData *responseData = [NSURLConnection sendSynchronousRequest:request returningResponse:&urlResponse error:&error];

        [self updateNetworkActivityIndicator:NO];

        if (error) {
            MixpanelError(@"%@ network failure: %@", self, error);
            break;
        }

        NSString *response = [[NSString alloc] initWithData:responseData encoding:NSUTF8StringEncoding];
        if ([response intValue] == 0) {
            MixpanelError(@"%@ %@ api rejected some items", self, endpoint);
        }

        [queue removeObjectsInArray:batch];
    }
}
```

This gets called on a serial background queue, so that part is fine. This tries to flush an `eventQueue` (just an array) every 60 seconds or when the app is backgrounded. If there's an `error` making the request, it doesn't remove the current batch (due to the `break`) and it will retry those the next time around.

Assuming there is no `error`, the Mixpanel client code here seems to expect a very primitive server response: `1` in case of success, `0` otherwise (0 is in fact only handled with a print, under the assumption that it's the Mixpanel server that decided that "some items should be rejected").

However, these seem like poor checks. There's a number of other recoverable failures that don't result in a non-nil `error` and will result in data loss.

Just one example: Say the user is in a coffee shop / on a hotel WiFi and the café/hotel runs a proxy through which the user must connect. If the app tries to send the request before the proxy allows it, then the `error` will be nil, but you will have sent the data to the proxy which will just discard it and the data is lost.

To simulate this, one thing you can do is just change the `serverURL` to something like `http://google.com`. Then what will happen is this:

```
2015-10-23 00:17:31.679 TestCenter[17879:4620107] <Mixpanel: 0x7f9982a39d30 2178e0df6f413afb8c43afe3bbd13f04> /track/ api rejected some items
<Breakpoint triggers at the end of the `flush` method>
(lldb) po error
 nil
(lldb) po response
<!DOCTYPE html>
… further stuff from Google …
```

While still not ideal, a simple patch that would not require server-side patches (and thus hopefully be backwards-compatible) would be:

```
diff --git a/Mixpanel/Mixpanel.m b/Mixpanel/Mixpanel.m
index 5d8a927..6e893ec 100755
--- a/Mixpanel/Mixpanel.m
+++ b/Mixpanel/Mixpanel.m
@@ -662,6 +662,13 @@ static __unused NSString *MPURLEncode(NSString *s)
         NSString *response = [[NSString alloc] initWithData:responseData encoding:NSUTF8StringEncoding];
         if ([response intValue] == 0) {
             MixpanelError(@"%@ %@ api rejected some items", self, endpoint);
+        } else if ([response intValue] == 1) {
+            // Success
+        } else {
+            // Another failure occurred. We expected a response of either `0` or `1`, but received something else.
+            // Maybe we were on hotel WiFi and had the request intercepted. Don't drop the data but retry again later.
+            // XXX Ideally we'd do something more intelligent, maybe by looking at `urlResponse` more closely.
+            break;
         }
 
         [queue removeObjectsInArray:batch];
```

Does that make sense? Is this something that should be patched? Just in case the answer is yes, I'll attach a PR.

dennda:
…hen traffic is being intercepted.

Fixes #369.

### DIFF
--- a/Mixpanel/Mixpanel.m
+++ b/Mixpanel/Mixpanel.m
@@ -662,6 +662,13 @@ static __unused NSString *MPURLEncode(NSString *s)
         NSString *response = [[NSString alloc] initWithData:responseData encoding:NSUTF8StringEncoding];
         if ([response intValue] == 0) {
             MixpanelError(@"%@ %@ api rejected some items", self, endpoint);
+        } else if ([response intValue] == 1) {
+            // Success
+        } else {
+            // Another failure occurred. We expected a response of either `0` or `1`, but received something else.
+            // Maybe we were on hotel WiFi and had the request intercepted. Don't drop the data but retry again later.
+            // XXX Ideally we'd do something more intelligent, maybe by looking at `urlResponse` more closely.
+            break;
         }
 
         [queue removeObjectsInArray:batch];


Mixpanel/Mixpanel.m
```diff
@@ -662,6 +662,13 @@ - (void)flushQueue:(NSMutableArray *)queue endpoint:(NSString *)endpoint
         NSString *response = [[NSString alloc] initWithData:responseData encoding:NSUTF8StringEncoding];
         if ([response intValue] == 0) {
             MixpanelError(@"%@ %@ api rejected some items", self, endpoint);
+        } else if ([response intValue] == 1) {
+            // Success
+        } else {
+            // Another failure occurred. We expected a response of either `0` or `1`, but received something else.
+            // Maybe we were on hotel WiFi and had the request intercepted. Don't drop the data but retry again later.
+            // XXX Ideally we'd do something more intelligent, maybe by looking at `urlResponse` more closely.
+            break;
```
dennda: @samgreen This of course assumes that the server doesn't ever send anything other than `0` or `1` (and that if it did, the request should be retried).

This is something I cannot know, but is something you'd have to verify with whoever runs that server.

samgreen: Can we change the loop to the invariant and break on `!= 1`? Prefer not to have the empty `if` there.

samgreen:
@dennda If you can make this one change we'll merge it in.

Thanks!

Mixpanel/Mixpanel.m
dennda: I am not sure exactly how you want to "change the loop". The loop is `while ([queue count] > 0)`, and you only have `response` later in its body (the request is performed synchronously). Also note that you *don't* want to break on `!= 1`, you want to break if `response.intValue != 0 && response.intValue != 1`.

Care to elaborate? :-)

poolqf: +1 to what @samgreen says...

```objc
        NSString *response = [[NSString alloc] initWithData:responseData encoding:NSUTF8StringEncoding];
        if ([response intValue] == 0) {
            MixpanelError(@"%@ %@ api rejected some items", self, endpoint);
        } else if ([response intValue] != 1) {
            // Another failure occurred. We expected a response of either `0` or `1`, but received something else.
            // Maybe we were on hotel WiFi and had the request intercepted. Don't drop the data but retry again later.
            // XXX Ideally we'd do something more intelligent, maybe by looking at `urlResponse` more closely.
            break;
        }
```

samgreen: Exactly! 

Can you change the if statement to check both of those cases?

PR #371, opened 2015-10-26
Option to suppress IP Address collection

idStar:
This feature lets client integrations decide on whether IP addresses should be sent along with other event tracking data. Normally, we want as much data as we can get, including the wonderful geo-location benefits possible with sending an IP address with events.

However, in some scenarios, our client app's privacy policies don't permit having the level of location specificity that is possible with an IP address.

As such, it is useful to be able to control whether we continue to transmit IP addresses (the default) or suppress doing so.

###### Usage Example

```
    // Initialize the library with our mixpanel token:
    [Mixpanel sharedInstanceWithToken:mixpanelToken];
    
    // Retrieve singleton instance of underlying analytics provider:
    self.analyticsProvider = [Mixpanel sharedInstance];
    self.analyticsProvider.transmitIPAddress = NO;
```

### DIFF
--- a/Mixpanel/Mixpanel.h
+++ b/Mixpanel/Mixpanel.h
@@ -189,6 +189,20 @@
 
 /*!
  @property
+ 
+ @abstract
+ Controls whether to automatically send the device's IP Address to
+ Mixpanel. Transmission allows Mixpanel to geo-locate a user down to
+ city-level specification. For privacy reasons, you may need to forego
+ having such collection.
+ 
+ @discussion
+ Defaults to YES.
+ */
+@property (atomic) BOOL transmitIPAddress;
+
+/*!
+ @property
 
  @abstract
  Determines the time, in seconds, that a mini notification will remain on
--- a/Mixpanel/Mixpanel.h
+++ b/Mixpanel/Mixpanel.h
@@ -193,14 +193,14 @@
  @abstract
  Controls whether to automatically send the client IP Address as part of 
  event tracking. With an IP address, geo-location is possible down to neighborhoods
- with a city, although the Mixpanel Dashboard will just show you city level location
+ within a city, although the Mixpanel Dashboard will just show you city level location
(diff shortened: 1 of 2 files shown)


idStar:
@samgreen Please let me know if there's something here you'd like handled differently so that the objective for this PR would be something you could incorporate. Thanks.

Mixpanel/Mixpanel.h
```diff
@@ -189,6 +189,21 @@
...
+ Controls whether to automatically send the client IP Address as part of 
+ event tracking. With an IP address, geo-location is possible down to neighborhoods
+ with a city, although the Mixpanel Dashboard will just show you city level location
+ specificity. For privacy reasons, you may be in a situation where you need to forego
+ effectively having access to such granular location information via the IP Address.
+ 
+ @discussion
+ Defaults to YES.
+ */
+@property (atomic) BOOL transmitIPAddress;
```
samgreen: Can we change this to `useIPAddressForGeoLocation`

samgreen:
Hi @idStar, this is a *fantastic* pull request. So sorry we didn't get a chance to merge it sooner. If we can just change the name of the `BOOL` we can merge it ASAP. Thank you for going the distance and writing docs as well.

idStar:
@samgreen Certainly. I'll make the changes you've requested and let you know when ready.

idStar:
@samgreen The change has been made. Here's an updated usage example:


###### Usage Example

```
    // Initialize the library with our mixpanel token:
    [Mixpanel sharedInstanceWithToken:mixpanelToken];
    
    // Retrieve singleton instance of underlying analytics provider:
    self.analyticsProvider = [Mixpanel sharedInstance];
    self.analyticsProvider.useIPAddressForGeoLocation = NO;
```

samgreen:
Thanks so much for your hard work!

PR #375, opened 2015-11-04
Added an unset: method to the Mixpanel People object

jaredmixpanel:
Added the ability to use the $unset operation to remove People properties from profiles


samgreen:
Thanks! @jaredmixpanel

PR #377, opened 2015-11-05
Documentation fixes in Mixpanel.h to fix iOS sample code

msintov:
1. Pass NSNumber* instead of int where required for this method:
- (void)increment:(NSString *)property by:(NSNumber *)amount;

2. Replace this sample code line (which doesn't compile):
[mixpanel.people identify:@"CURRENT USER DISTINCT ID"];

...with this line:
[mixpanel identify:@"CURRENT USER DISTINCT ID"];

3. Remove this sample code line that doesn't compile:
[mixpanel.people identify:distinctId];


samgreen:
Looks great, thanks!

PR #380, opened 2015-11-10
tvos 9.0 support

pantuspavel:
Maybe it even make sense to add watchOS support as well.


pantuspavel:
Looks like mixpanel is not ready yet (a lot of warnings and errors). Looking forward to tracking events on tvOS. In the meantime closing PR.

rodericj:
@pantuspavel yes it seems as though the SDK now includes WatchConnectivity which currently breaks a build for tvOS. I'm wondering if it is on the roadmap to address this.

PR #387, opened 2015-12-01
Flush events after setting flush interval

patedit:
If the flush interval is dynamically changing, we could end up in a situation where no events are sent during the user session.


samgreen:
Looks good to me.

PR #389, opened 2015-12-01
Make sure all background tasks are killed when expired

patedit:
If the app is background/foregrounded very quickly, we can't guarantee that we end all the tasks after they expire since self.taskId keeps a reference to the last executed task on the background.


samgreen:
:+1:

PR #392, opened 2015-12-09
Protect against nil dictionary access

rroy1590:
Our production app has been seeing crashes in the wild in these two Mixpanel functions. Both crash logs indicate invalid subscript access on the dictionary which led us to believe the mutableCopy failed to return a valid object for some reason.
screen shot 2015-12-08 at 4 32 06 pm
screen shot 2015-12-08 at 4 31 57 pm

### DIFF
--- a/Mixpanel/Mixpanel.m
+++ b/Mixpanel/Mixpanel.m
@@ -921,8 +921,10 @@ static __unused NSString *MPURLEncode(NSString *s)
 {
     dispatch_async(self.serialQueue, ^(){
         NSMutableDictionary *properties = [self.automaticProperties mutableCopy];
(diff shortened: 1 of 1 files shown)


samgreen:
Strange that you're seeing issues with this. We definitely call `collectAutomaticProperties` which should populate that dictionary before we call `setUpListeners` which would call `setCurrentRadio`.

Regardless, if you've been seeing crashes there is definitely something going on. I'm all for more defensive programming, so thanks for the PR!

Please don't hesitate to submit further PRs if you find other issues with the library.

rroy1590:
@samgreen thanks for responding so quickly!

PR #396, opened 2015-12-15
more efficient system version compare

Coeur:


samgreen:
Hey @Coeur Can you please change the magic numbers back to the constants? We only support iOS 8+ so there is no concern about the constant being unavailable. Happy to merge with that change!

Coeur:
@samgreen, those constants are not dependent on target OS, but on Xcode version. So there is a concern at least for `NSFoundationVersionNumber_iOS_8_0`:
* it is not available on Xcode 6.0
* Xcode 6.0 supports iOS 8 and upload to iTunes Connect

But true, if you only support iOS 8+, we can safely use NSFoundationVersionNumber_iOS_7_0.

Coeur:
@samgreen Sorry, but if you really only support iOS 8+, we can remove the version compare to iOS 7 altogether, right?

samgreen:
This is a recent change so the podspec should be updated. Running the execellent, Deploymate, shows this:

image

samgreen:
We should remove the version checks for clarity.

Coeur:
Can I provide fixes for what Deploymate found, so we can continue to support iOS 7 or iOS 6, or is it a business decision and there is no going back?

My app is supporting iOS 7, so I would not be able to use latest Mixpanel anymore if you move to iOS 8.

Coeur:
I've checked with Deploymate: it finds what is already enclosed in iOS version conditions!

    if (device is running iOS 8) {
        [iOS 8+ feature];
    } else {
        [iOS 2+ fallback];
    }

Deploymate is warning you for the *[iOS 8+ feature]* being not available on iOS 7.

samgreen:
Sorry @Coeur I should have read the release notes more clearly. In the future we are supporting iOS 8+ with Cocoapods and Carthage. iOS 7 integrations will have to be done manually.

Coeur:
Please attempt to do like famous AFNetworking Cocoapods support:
* AFNetworking release branches 1.x for iOS 5+
* AFNetworking release branches 2.x for iOS 6+
* AFNetworking release branches 3.x for iOS 7+
* AFNetworking release branches 4.x for iOS 8+

It could be:
* Mixpanel release branches 2.x for iOS 6+ (current podspec)
* Mixpanel release branches 3.x for iOS 8+ (apparently what you want)

samgreen:
Thanks!

PR #399, opened 2015-12-22
Nullability audit on MPTweak, Mixpanel, and MixpanelPeople.

samgreen:
This will make interoperability much better for Swift developers, and require less use of Optional types.


patedit:
Looks good!

PR #403, opened 2015-12-23
Resolves the casting warning received in NSData+MPBase64.m

samgreen:


Coeur:
This was only an issue in Xcode 6. Using Xcode 7, the good practice is the opposite: removing the superfluous casting.

PR #407, opened 2015-12-31
On reset, initialize variants and eventBindings

peymano:
These should be reset as well, otherwise, they're unarchived from a previous instance/project.


RIAlizer:
+1

samgreen:
Thanks!

PR #410, opened 2016-01-05
avoiding casting options

Coeur:


Quuxplusone:
FWIW, StackOverflow seems to think that `kNilOptions` is "un-modern" style and should generally be avoided. I have no experience with it but would tend to agree. Does Xcode complain about C-style casts now? and if so, why not just remove the casts?
http://stackoverflow.com/questions/24758086/nil-vs-kniloptions

samgreen:
I prefer using `0`, we do not use `kNilOptions` anywhere, and it is indeed ancient.

Coeur:
Only few votes on StackOverflow for that one.
@Quuxplusone, removing the casts causes warnings (NSInteger vs Enum value), so you can't remove the casts
@samgreen, sure, if you prefer casting, it is up to you and you can leave things as they are; @yas375 and I prefer `kNilOptions` which is an anonymous Enum for the purpose of avoiding casting.

samgreen:
Not merging this to keep our code consistent.

PR #413, opened 2016-01-15
Prevent negative $duration values

jaredmixpanel:
Fixes a race condition between timeEvent: and track: that leads to negative $duration values, due to the start time being captured asynchronously, while the event timestamp is recorded on the main thread. To prevent the start time from being evaluated after the event has been tracked, we simply capture the start time inside of timeEvent: before dispatching asynchronously.


samgreen:
Thanks!

PR #419, opened 2016-01-21
Add functionality to unset a user's push device token.

### LINKED ISSUE #164: Unset Device Tokens
I have a use case for "unsetting" device tokens via the iOS sdk. We have log-in and log-out functionality in our app. It's possible for a user to log-in with account X, log-out of that account, and then log in with account Y (all on the same device). 

I would like to be able to unregister the deviceToken for account X when they log-out.

samgreen:
Resolves #164

### DIFF
--- a/Mixpanel/Mixpanel.h
+++ b/Mixpanel/Mixpanel.h
@@ -803,6 +803,18 @@ NS_ASSUME_NONNULL_BEGIN
 
 /*!
  @method
+ 
+ @abstract
+ Unregister the given device to receive push notifications.
+ 
+ @discussion
+ This will unset all of the push tokens saved to this people profile. This is useful
+ in conjunction with a call to `reset`, or when a user is logging out.
+ */
+- (void)removePushDeviceToken;
+
+/*!
+ @method
 
  @abstract
  Set properties on the current user in Mixpanel People.
--- a/Mixpanel/Mixpanel.m
+++ b/Mixpanel/Mixpanel.m
@@ -1969,6 +1969,12 @@ static void MixpanelReachabilityCallback(SCNetworkReachabilityRef target, SCNetw
     [self addPeopleRecordToQueueWithAction:@"$union" andProperties:properties];
 }
 
+- (void)removePushDeviceToken
+{
+    NSDictionary *properties = @{ @"$properties": @"$ios_devices" };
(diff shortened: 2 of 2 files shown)


patedit:
Looks good!

cerupcat:
@samgreen Doesn't this just clear all of the tokens? We're looking to remove a specific token. As reason would be that the token is no longer valid. Other tokens in the array may still be valid (so we don't want to clear all of them).

samgreen:
@cerupcat yes it does. I guess I misunderstood your issue. Would something like removePushDeviceToke:actualToken be more helpful?

cerupcat:
@samgreen Yeah, the ability to remove a specific token is what we're looking for (to be able to remove invalid ones). It doesn't appear there's a way (that we know of) to do this today.

samgreen:
@cerupcat There is no documented method available to /engage/ that enables this behavior. We are investigating a solution for you.

b1az:
@samgreen Is perhaps a solution for removing token per device already in the making?

samwize:
@samgreen Hi. Is there an update on this issue? 

We also need the method to clear only the token for the current device, not all devices for a user.

cerupcat:
Yeah, right now on iOS9 if a user uninstalls and reinstalls an app, they'll get a new token, but Apple doesn't invalidate the previous. Since the new token would be sent to Mixpanel, a user will end up getting multiple notifications on that device. It'd be nice if Mixpanel stored only one token per device or there was a way of clearing tokens (on first app launch or something).

samgreen:
@cerupcat this is not technically true. When the app is uninstalled, the push token becomes invalid and the user can no longer receive push notifications from that token. You have always had the ability to clear all tokens with `$unset`.

However, we have heard you loud and clear. We have a previously undocumented `$remove` people operation that we were waiting on sign off from our systems team to make use of. You can see the details [here](https://mixpanel.com/help/reference/http). We received the 👍  from our systems team, and I'm pleased to let you know that we've implemented this change in #554 

You will have this feature in the next release of our SDK. Thanks for your patience!

cerupcat:
Thanks @samgreen, looking forward to it.

I can't find the issue, but on iOS9 (can't recall what version(s)) there's a known issue where Apple doesn't invalidate the token, so you can actually get multiple sent (even though you have a new one).

Looking forward to check out the next release. :)

PR #421, opened 2016-01-26
Added support for watchOS 2

samgreen:


Mixpanel-watchOS-Info.plist
```diff
@@ -0,0 +1,24 @@
...
+	<string>$(PRODUCT_NAME)</string>
+	<key>CFBundlePackageType</key>
+	<string>FMWK</string>
+	<key>CFBundleShortVersionString</key>
+	<string>1.0</string>
+	<key>CFBundleSignature</key>
+	<string>????</string>
+	<key>CFBundleVersion</key>
+	<string>$(CURRENT_PROJECT_VERSION)</string>
+	<key>NSPrincipalClass</key>
```
patedit: ?

Mixpanel/Mixpanel.m
```diff
@@ -1854,8 +1852,33 @@ - (void)executeCachedEventBindings {
...
+    if ([Mixpanel isValidWatchSessionMessage:message]) {
+        [[Mixpanel sharedInstance] track:message[@"event"] properties:message[@"properties"]];
+        replyHandler(@{ @"success": @YES });
+    } else {
+        replyHandler(@{ @"success": @NO, @"message": @"Message is not a mixpanel message" });
+    }
+}
+
++ (BOOL)isValidWatchSessionMessage:(NSDictionary<NSString *, id> *)message {
+    return [[message objectForKey:@"$mp_message_type"] boolValue];
```
patedit: We just need to check if `$mp_message_type` is contained in the dictionary. `boolValue` will cause issues if we are passing a number, for example.

Mixpanel/MixpanelWatchOS.m
```diff
@@ -0,0 +1,74 @@
...
+}
+
+- (void)track:(NSString *)event {
+    [self track:event properties:nil];
+}
+
+- (void)track:(NSString *)event properties:(nullable NSDictionary *)properties {
+    NSAssert(event != nil, @"Missing event name");
+    
+    if ([self.session isReachable]) {
```
patedit: Let's externalize this to another method. `track` will just prepare the data and send a `$mp_message_type` of `track` to this new method. Then we will parse each `$mp_message_type` string on the `session:didReceiveMessage` and act accordingly (for now, only track events)

Mixpanel/Mixpanel.m
```diff
@@ -1854,8 +1852,33 @@ - (void)executeCachedEventBindings {
...
 @end
 
-#pragma mark - People
+#pragma mark - WatchExtensions
+@implementation Mixpanel (WatchExtensions)
+
+/** Called on the delegate of the receiver. Will be called on startup if the incoming message caused the receiver to launch. */
+- (void)session:(WCSession *)session didReceiveMessage:(NSDictionary<NSString *, id> *)message {
+    if ([Mixpanel isValidWatchSessionMessage:message]) {
+        [[Mixpanel sharedInstance] track:message[@"event"] properties:message[@"properties"]];
```
patedit: We will have to parse the message[@"$mp_message_type"] type here and act accordingly (i.e. track, increment, etc..). We will only support track for now.

samgreen:
It's correct for building with Carthage.

PR #424, opened 2016-01-27
Networking hardening for the SDK

samgreen:


Mixpanel/Mixpanel.m
```diff
@@ -693,6 +702,39 @@ - (NSURLRequest *)apiRequestWithEndpoint:(NSString *)endpoint andBody:(NSString
     return request;
 }
 
+- (void)handleNetworkResponse:(NSHTTPURLResponse *)response withError:(NSError *)error
+{
+    NSNumber *retryTime = response.allHeaderFields[@"Retry-After"];
+    
+    BOOL was5XX = (response.statusCode >= 500 && response.statusCode < 600);
```
Quuxplusone: might read clearer as `(500 <= ... && ... <= 599)`

Mixpanel/Mixpanel.m
```diff
@@ -693,6 +702,39 @@ - (NSURLRequest *)apiRequestWithEndpoint:(NSString *)endpoint andBody:(NSString
...
+        }
+    }
+    
+    self.networkRequestsAllowedAfterTime = [[NSDate dateWithTimeIntervalSinceNow:retryTime.doubleValue] timeIntervalSince1970];
+}
+
+- (NSNumber *)retryBackOffTime
+{
+    const NSUInteger retryInterval = (arc4random() % 30) + 30;
+    return @(pow(retryInterval, self.networkConsecutiveFailures));
```
Quuxplusone: you want this to be more like

    pow(2.0, self.networkConsecutiveFailures) * 60 + randint(30)

Also, that `2.0` could be lowered to `1.5` if we wanted; it's a parameter to the algorithm.

Quuxplusone: in fact you could simplify this whole business to

    retryTime = min(max(60, [self retryBackOffTime]), 600)

Mixpanel/Mixpanel.m
```diff
@@ -651,6 +654,11 @@ - (void)flushPeople
 
 - (void)flushQueue:(NSMutableArray *)queue endpoint:(NSString *)endpoint
 {
+    if (self.networkRequestsAllowedAfterTime > [[NSDate date] timeIntervalSince1970]) {
```
Quuxplusone: reverse operands and `<`

Mixpanel/Mixpanel.m
```diff
@@ -147,6 +149,7 @@ - (instancetype)initWithToken:(NSString *)apiToken launchOptions:(NSDictionary *
         [[MixpanelExceptionHandler sharedHandler] addMixpanelInstance:self];
 #endif
         
+        self.networkRequestsAllowedAfterTime = -1;
```
Quuxplusone: `s/-1/0/`

Mixpanel/Mixpanel.m
```diff
@@ -693,6 +702,31 @@ - (NSURLRequest *)apiRequestWithEndpoint:(NSString *)endpoint andBody:(NSString
     return request;
 }
 
+- (void)handleNetworkResponse:(NSHTTPURLResponse *)response withError:(NSError *)error
+{
+    NSNumber *retryTime = response.allHeaderFields[@"Retry-After"];
```
patedit: We're not using the header information at all. If this field is present we should pick this time if it's larger than `pow(2.0, failureCount) * 60 + arc4random_uniform(30)`.

Mixpanel/Mixpanel.m
```diff
@@ -693,6 +702,31 @@ - (NSURLRequest *)apiRequestWithEndpoint:(NSString *)endpoint andBody:(NSString
...
+        retryTime = [self retryBackOffTimeWithConsecutiveFailures:self.networkConsecutiveFailures];
+    }
+    
+    self.networkRequestsAllowedAfterTime = [[NSDate dateWithTimeIntervalSinceNow:retryTime.doubleValue] timeIntervalSince1970];
+}
+
+- (NSNumber *)retryBackOffTimeWithConsecutiveFailures:(NSUInteger)failureCount
+{
+    NSTimeInterval time = pow(2.0, failureCount) * 60 + arc4random_uniform(30);
+    return @(MIN(MAX(60, time), 600));
```
patedit: Instead of 60, shouldn't we use `flushInterval` instead?

Issue #329, opened 2015-08-31
A question about using cache when I call Mixpanel#showNotification


yas375:
We have an issue with Mixpanel which we have to workaround in our fork of the library. Thank you guys for making the library open source! :heart:

I'd like to explain why and what issue do we have and how we workaround it. Maybe you'll see that it makes sense to make a change to the library itself. BTW in this case I'm more then happy to submit a pull request if needed ;)

---

While user uses the app we log ananlytic events and properties to Mixpanel. We have 60 seconds flush interval.

In Mixpanel we target users with particular notifications based on analytics data we have tracked while user uses the app.

<img width="709" alt="2015-08-30_2229" src="https://cloud.githubusercontent.com/assets/217368/9572620/a86dc384-4f66-11e5-81fe-8070d58007d5.png">

We have multiple places in the app where we trigger Mixpanel to present a notification if there is one. With this repo's implementation it will be 

```objective-c
- (void)showNotification
{
    [self checkForNotificationsWithCompletion:^(NSArray *notifications) {
        if ([notifications count] > 0) {
            [self showNotificationWithObject:notifications[0]];
        }
    }];
}
```

Where `checkForNotificationsWithCompletion:` calls `checkForDecideResponseWithCompletion:` which results in `[self checkForDecideResponseWithCompletion:completion useCache:YES];` call. As a result inside this method we don't enter inside [the if state](https://github.com/mixpanel/mixpanel-iphone/blob/7d463a07ae585330dd7ba65ed22f1ccbe532dda4/Mixpanel/Mixpanel.m#L1213) and don't fetch notifications from Mixpanel's server. 

```
- (void)checkForDecideResponseWithCompletion:(void (^)(NSArray *surveys, NSArray *notifications, NSSet *variants, NSSet *eventBindings))completion useCache:(BOOL)useCache
{
    dispatch_async(self.serialQueue, ^{
        ...
        if (!useCache || !self.decideResponseCached) {
            ...
        }
        ....
    });
}
```

The problem is that based on the most recently tracked data if the SDK will send the request it will get a notification for the user, but it doesn't send the request because of `useCache:YES` and `self.decideResponseCached` becomes `NO` only in a few cases (the app enters background or `Mixpanel#reset` is called).

We workaround it by adding `showNotificationUsingCache:` method: https://github.com/carezone/mixpanel-iphone/commit/ec0509118aec9bc67a6fdeaa9da3601a9b7dff22

[...]

PR #426, opened 2016-01-28
Don't use the cache internally when the developer calls checkForSurvey: or checkForNotification:

samgreen:


Mixpanel/Mixpanel.m
```diff
@@ -1413,7 +1413,7 @@ - (void)checkForNotificationsWithCompletion:(void (^)(NSArray *notifications))co
         if (completion) {
             completion(notifications);
         }
-    }];
+    } useCache:NO];
```
yas375: :+1: It will close #329 :)

PR #430, opened 2016-02-04
Added signalHandler to Mixpanel 

uzysjung:
UncaughtExceptionHandler can not catch specific crash such as segmantation fault, floating point exception, bus error, etc.

In order to prevent to lose data from this crash, use signalHandler.

I added SignalHandler.
It prevent to lose data when app is crashed.

Thanks.

### DIFF
--- a/Mixpanel/MixpanelExceptionHandler.m
+++ b/Mixpanel/MixpanelExceptionHandler.m
@@ -10,6 +10,8 @@
 #import "Mixpanel.h"
 #import "MPLogger.h"
 
+#include <libkern/OSAtomic.h>
+
 @interface MixpanelExceptionHandler ()
 
 @property (nonatomic) NSUncaughtExceptionHandler *defaultExceptionHandler;
@@ -18,6 +20,19 @@
 @end
 
 @implementation MixpanelExceptionHandler
+
+static uint32_t volatile isAlreadyExceptionOccured =0;
+static int fatal_signals[] =
+{
+    SIGILL  ,   /* illegal instruction (not reset when caught) */
+    SIGTRAP ,   /* trace trap (not reset when caught) */
+    SIGABRT ,   /* abort() */
+    SIGFPE  ,   /* floating point exception */
+    SIGBUS  ,   /* bus error */
+    SIGSEGV ,   /* segmentation violation */
+    SIGSYS  ,   /* bad argument to system call */
+};
+static int n_fatal_signals = (sizeof(fatal_signals) / sizeof(fatal_signals[0]));
 
 + (instancetype)sharedHandler {
     static MixpanelExceptionHandler *gSharedHandler = nil;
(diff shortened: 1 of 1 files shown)


Mixpanel/MixpanelExceptionHandler.m
```diff
@@ -33,38 +48,86 @@ - (instancetype)init {
...
 
 static void mp_handleUncaughtException(NSException *exception) {
     MixpanelExceptionHandler *handler = [MixpanelExceptionHandler sharedHandler];
-    
-    // Archive the values for each Mixpanel instance
-    for (Mixpanel *instance in handler.mixpanelInstances) {
-        // Since we're storing the instances in a weak table, we need to ensure the pointer hasn't become nil
-        if (instance) {
-            [instance archive];
+    NSLog(@"!!!!!!");
```
samgreen: Can you please remove this?

Mixpanel/MixpanelExceptionHandler.m
```diff
@@ -33,38 +48,86 @@ - (instancetype)init {
...
+    if(isAlreadyExceptionOccured==0) {
```
samgreen: 1 space after keywords `if (`

Mixpanel/MixpanelExceptionHandler.m
```diff
@@ -33,38 +48,86 @@ - (instancetype)init {
...
     if (handler.defaultExceptionHandler) {
         // Ensure the existing handler gets called once we're finished
         handler.defaultExceptionHandler(exception);
     }
 }
 
+static void mp_handleSignal(int sig, siginfo_t *info, void *context) {
+    unregisterFatalSignals();
+    MixpanelError(@"We received a signal: %d", sig);
+    NSLog(@"??????");
```
samgreen: Remove this please

Mixpanel/MixpanelExceptionHandler.m
```diff
@@ -33,38 +48,86 @@ - (instancetype)init {
...
+
+    NSException* exception = [NSException
+                              exceptionWithName:@"UncaughtException"
+                              reason:
+                              [NSString stringWithFormat:
+                               NSLocalizedString(@"Signal %d was raised.", nil),
```
samgreen: Please break after the first colon and align other colons. Xcode can do this with `Control + I`

Mixpanel/MixpanelExceptionHandler.m
```diff
@@ -33,38 +48,86 @@ - (instancetype)init {
...
+
+static void registerFatalSignals() {
+    struct sigaction sa;
+    /* Configure action */
+    memset(&sa, 0, sizeof(sa));
+    sa.sa_flags =  SA_SIGINFO | SA_ONSTACK;
+    sa.sa_sigaction = &mp_handleSignal;
+    sigemptyset(&sa.sa_mask);
+    /* Set new sigaction */
+    for (int i =0 ;i<n_fatal_signals; i++) {
```
samgreen: Spacing here. Should be: 

`for (int i =0 ;  i< n_fatal_signals; i++) {`

Mixpanel/MixpanelExceptionHandler.m
```diff
@@ -33,38 +48,86 @@ - (instancetype)init {
...
+        if (sigaction(fatal_signals[i], &sa, NULL) != 0) {
+            //            int err = errno;
+            //            NSAssert(0,"Signal registration for %s failed: %s", strsignal(fatal_signals[i]), strerror(err));
```
samgreen: No action here? In this case we should remove the empty case.

samgreen:
Will review based on content tomorrow.

Anyway you could provide some basic unit tests for this functionality?

uzysjung:
Hi I fixed what you requested and I can provide unit test.
But All signalhandler is composed of private method and static c function. 
In order to make unit test. Use method swizzling And change sourcecode.
If you need it plz let me know.

uzysjung:
you can test after disable debugger. 

use code below 

1. CFRelease(NULL);
2. [self performSelector:@selector(string) withObject:nil afterDelay:2.0];

3. void (*nullFunction)() = NULL; nullFunction();

samgreen:
Upon further review, I do not believe we should add these other handlers to the Mixpanel SDK.

Here are my concerns taken from [Mike Ash's](https://www.mikeash.com/pyblog/friday-qa-2011-04-01-signal-handling.html) thoughts on this:
> You can call your own reentrant code, but you probably don't have any, because it's hard to write, it can't call any system functions except from the above list, and you never had any reason to write it before. For the Objective-C types, note that objc_msgSend is not reentrant, so you cannot use any Objective-C from a signal handler.

Since we can't call `objc_msgSend` we *cannot* call `archive` safely from within the signal handler.

> There is very little that you can do safely. There is so little that I'm not even going to discuss how to get anything done, because it's so impractical to do so, and instead will simply tell you to avoid using signal handlers unless you really know what you're doing and you enjoy pain.

I would prefer to do this using a `kqueue` or even better using GCD.

> The best way to handle a signal in almost every case is to use GCD. Signal handling with GCD is easy and safe. On the rare occasions where you need to handle signals, GCD lets you do it with just a few lines of code.

Happy to review a second PR that implements this in a safer manner.

uzysjung:
you are right 
FYI
https://www.securecoding.cert.org/confluence/display/c/SIG30-C.+Call+only+asynchronous-safe+functions+within+signal+handlers

PR #432, opened 2016-02-09
Remove team id from xcodeproj

izqui:
Carthage build was failing because of that:

```
Code Sign error: No code signing identities found: No valid signing identities (i.e. certificate and private key pair) matching the team ID “E8FVX7QLET” were found.
```


patedit:
Thanks @izqui !

Issue #429, opened 2016-02-04
A/B experiments persist even when turned off and backgrounded, until the app is re-launched


cassiegamm:
**Outstanding issue:**
The app often requires TWO re-launches before reverting to default values (aka disabling an experiment). This makes it difficult to pull back on features/experiments we don’t want users to see anymore (bugs for example).

**Lack of Parity:**
I think from an engineering perspective parity is super important to keep things clear. Why should changes in variant distribution be applied immediately upon didBecomeActive but changing the experiment from ON to OFF not be applied until after the app is closed and re-launched? (sometime more than once). 

The ability to turn off an experiment quickly is important in the case of major unpredicted bugs, or a business externality.  App relaunch is much slower than app active.

Ideally, engineers should have the option of customizing the behavior that applies (or removes!) experiments during didBecomeActive.

ryanseams:
i wanted to shed some additional light on this issue with the results from some testing.

i found that when a test is either: 1) turned off OR 2) the variant allocation is changed, you must make a call to ```[mixpanel checkForDecideResponseWithCompletion:completion useCache:NO];``` in order to force the SDK to get an update from Mixpanel. in this update, you will receive a new response from the /decide server based on the changes you make in the web UI.

at the point of receiving and applying the experiment (the cache clearing and experiment application all happens by default in ```joinExperimentsWithCallback:```), two things happen: 1) any UI changes are discarded from the prior experiment AND 2) tweaks remain applied from the prior experiment, not the new one.

i would expect, given the behavior says useCache:NO, that we simply discard the old A/B test and apply whatever is received (a new variant, or the original app experience). i believe the issue here is tweak values continue to persist, which appears to be some sort of issue with how the SDK manages the variants [here](https://github.com/mixpanel/mixpanel-iphone/blob/master/Mixpanel/Mixpanel.m#L1324-L1340).

let me know if i can assist further or provide any additional context here as needed.

samgreen:
Resolved with #442

PR #442, opened 2016-02-16
Apply new variants / Turn off experiments when calling joinExperimentsWithCallback

ryanseams:
Currently the SDK will check for new experiments via ```joinExperimentsWithCallback``` and process the updates received from Mixpanel servers. When a new variant is received or no variants are received, the running variant is [marked](https://github.com/mixpanel/mixpanel-iphone/blob/master/Mixpanel/Mixpanel.m#L1376-L1377) under the set ```toFinishVariants``` as it should be.

The issue is that for these finished variants, we call to [finish](https://github.com/mixpanel/mixpanel-iphone/blob/master/Mixpanel/MPVariant.m#L231-L233) which does not actually end the variant shown to the user. Instead, we should be calling [stop](https://github.com/mixpanel/mixpanel-iphone/blob/master/Mixpanel/MPVariant.m#L221-L229) to stop running this variant and allow the new changes from Mixpanel to be applied.

This resolves both cases highlighted in https://github.com/mixpanel/mixpanel-iphone/issues/429:

1. Receiving a new variant and applying the changes
2. Receiving no variant and turning the experiment off (reverting values to defaults)

Since the SDK will only check for new experiments when using ```joinExperimentsWithCallback```, updates from Mixpanel will only be applied when this method is called (which is the expected user behavior).

### DIFF
--- a/Mixpanel/Mixpanel.m
+++ b/Mixpanel/Mixpanel.m
@@ -1387,7 +1387,7 @@ static void MixpanelReachabilityCallback(SCNetworkReachabilityRef target, SCNetw
             [allVariants unionSet:runningVariants];
 
             [restartVariants makeObjectsPerformSelector:NSSelectorFromString(@"restart")];
-            [toFinishVariants makeObjectsPerformSelector:NSSelectorFromString(@"finish")];
+            [toFinishVariants makeObjectsPerformSelector:NSSelectorFromString(@"stop")];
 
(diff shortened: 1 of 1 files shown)


samgreen:
Thanks!

rowrow:
Hi MP team, 
Do you have plans to re-fix this? Without it, we have to call reset() in weird places and it's not the best practice. Thanks!

PR #445, opened 2016-02-16
New target 'Mixpanel_appExtension' for carthage to use app extensions. Added 'Require Only App-Extension-Safe API' to target.

### LINKED ISSUE #394: Extension API warnings when installed via Carthage
I'm using Carthage to add Mixpanel to my project. When do I do this I get xcode warnings "Linking against dylib not safe for use in application extensions: /Users/blah/Mixpanel.framework/Mixpanel"

samgreen:
Resolves #394

### DIFF
--- a/Mixpanel-AppExtension-Info.plist
+++ b/Mixpanel-AppExtension-Info.plist
@@ -5,7 +5,7 @@
 	<key>CFBundleDevelopmentRegion</key>
 	<string>en</string>
 	<key>CFBundleExecutable</key>
-	<string>Mixpanel</string>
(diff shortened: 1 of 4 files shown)


JamieWhite:
So I've been playing around with this but I don't think it actually works. I'm trying to import the new MixpanelAppExtension framework into my project and xcode keeps failing to build.

can you confirm this actually works in a test project?

raheelahmad:
Ditto what @JamieWhite said. Does importing `MixpanelAppExtension` actually work?

PR #459, opened 2016-04-01
Fix for watchOS integration

### LINKED ISSUE #449: Want to use mixpanel on ios app & extension as well as watchos
v2.9.3 podspec seems broken a bit - it lacks for AppExtension subspec.
v2.9.4 isn't accessible (actually pod gives me just v2.9.1, not later).
I can't wait to get extension spec repaired + watchos support.

samgreen:
Resolves watchOS integration when using more than one instance of the Mixpanel SDK. Resolves #449.


patedit:
Looks good :+1:

PR #464, opened 2016-04-01
Archive on every flush

### LINKED ISSUE #441: Question about flushWithCompletion
In:

```
- (void)flushWithCompletion:(void (^)())handler
``` 
there is this code:

```
if (handler) {
    [self archive];
    dispatch_async(dispatch_get_main_queue(), handler);
}
```

I am wondering, what is the reason for archive being called only if handler is passed?

samgreen:
Call archive on every flush, regardless of whether a completion block is passed.

Resolves #441


patedit:
Much needed change :+1:

PR #469, opened 2016-04-09
Added tldr CocoaPods: pod 'Mixpanel'.

orarbel:


yas375:
Nit-picking about commit's message: it's called CocoaPods ;)

Issue #420, opened 2016-01-21
Adding failure block to `joinExperimentsWithCallback:`?


yas375:
## Current state

[Current API](https://github.com/mixpanel/mixpanel-iphone/blob/2106d67e4197b3244f202793d54d6e580152773e/Mixpanel/Mixpanel.h#L735):

```objective-c
- (void)joinExperimentsWithCallback:(nullable void (^)())experimentsLoadedCallback;
```

The callback is executed only in case everything was loaded successfully. For example if there is an [error with your service](https://twitter.com/mixpanel/status/689990498518458369) and it returns an error, then it just returns [here](https://github.com/mixpanel/mixpanel-iphone/blob/2106d67e4197b3244f202793d54d6e580152773e/Mixpanel/Mixpanel.m#L1278) and callback won't be called.

## The problem

I agree that in a lot of cases people doesn't need to know about a failure. But there might be a case when  we need. For example we want to test different navigation types in an app. To do this, we need to fetch variants at the end of sign in sign up before actually dismissing the sign up screen. We are fine to spin a spinner for some time while variants are fetched from Mixpanel's server. Was it successful or not - we need to dismiss our sign up screen. In case of a failure fetching variants we can use some default values, but we still need to dismiss the sign up screen. Currently if we do something like this:

```objective-c
  [self.mixpanel joinExperimentsWithCallback:^{
    // dismiss sign up screen
  }];
```

we are not getting guaranteed execution of the block.

## Proposed solution

Additionally to existent `joinExperimentsWithCallback:` we can add an additional method which will also take a failure callback:

```objective-c
- (void)joinExperimentsWithCallback:(nullable void (^)())experimentsLoadedCallback;
- (void)joinExperimentsWithCallback:(nullable void (^)())experimentsLoadedCallback onFailure:(nullable void (^)(NSError *))failureCallback;
```

Original method will call the new one passing `nil` as `failureBlock`.

This will also require similar modifications to `checkForVariantsWithCompletion:` and `checkForDecideResponseWithCompletion:useCache:`. It will by 100% backward compatible.

---

I'm happy to volunteer to implement this and submit a pull request. What do you think?

samgreen:
Love it, and would be happy to merge a PR today or next week.

samgreen:
Massive oversight on our part never calling the completion block.

peymano:
[...]

PR #470, opened 2016-04-13
Make sure completion handler is called even when a decide check returns an error

jaredmixpanel:
Currently, if we get an error back from decide in `checkForDecideResponseWithCompletion:completion useCache:` we log it and return without calling the completion handler. This prevents `joinExperimentsWithCallback:` from ever calling its callback block. This fix simply calls the completion handler (if it exists) with `nil` values in the event of a decide error.

### DIFF
--- a/Mixpanel/Mixpanel.m
+++ b/Mixpanel/Mixpanel.m
@@ -1337,7 +1337,7 @@ static void MixpanelReachabilityCallback(SCNetworkReachabilityRef target, SCNetw
             if (error) {
                 MixpanelError(@"%@ decide check http error: %@", self, error);
                 if (completion) {
-                    completion(nil,nil,nil,nil);
+                    completion(nil, nil, nil, nil);
                 }
                 return;
             }
@@ -1345,14 +1345,14 @@ static void MixpanelReachabilityCallback(SCNetworkReachabilityRef target, SCNetw
             if (error) {
                 MixpanelError(@"%@ decide check json error: %@, data: %@", self, error, [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding]);
                 if (completion) {
-                    completion(nil,nil,nil,nil);
+                    completion(nil, nil, nil, nil);
                 }
                 return;
             }
             if (object[@"error"]) {
                 MixpanelDebug(@"%@ decide check api error: %@", self, object[@"error"]);
                 if (completion) {
-                    completion(nil,nil,nil,nil);
+                    completion(nil, nil, nil, nil);
                 }
                 return;
             }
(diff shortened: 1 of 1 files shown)


samgreen:
Hey @jaredmixpanel thanks for another PR buddy!

Does this resolve #420? I'm thinking it does. I have some minor style gripes and then we can merge.

Mixpanel/Mixpanel.m
```diff
@@ -1336,15 +1336,24 @@ - (void)checkForDecideResponseWithCompletion:(void (^)(NSArray *surveys, NSArray
             NSData *data = [NSURLConnection sendSynchronousRequest:request returningResponse:&urlResponse error:&error];
             if (error) {
                 MixpanelError(@"%@ decide check http error: %@", self, error);
+                if (completion) {
+                    completion(nil,nil,nil,nil);
```
samgreen: Should be `completion(nil, nil, nil, nil);`

Mixpanel/Mixpanel.m
```diff
@@ -1336,15 +1336,24 @@ - (void)checkForDecideResponseWithCompletion:(void (^)(NSArray *surveys, NSArray
...
+                }
                 return;
             }
             NSDictionary *object = [NSJSONSerialization JSONObjectWithData:data options:(NSJSONReadingOptions)0 error:&error];
             if (error) {
                 MixpanelError(@"%@ decide check json error: %@, data: %@", self, error, [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding]);
+                if (completion) {
+                    completion(nil,nil,nil,nil);
```
samgreen: Should be `completion(nil, nil, nil, nil);`

Mixpanel/Mixpanel.m
```diff
@@ -1336,15 +1336,24 @@ - (void)checkForDecideResponseWithCompletion:(void (^)(NSArray *surveys, NSArray
...
+                }
                 return;
             }
             if (object[@"error"]) {
                 MixpanelDebug(@"%@ decide check api error: %@", self, object[@"error"]);
+                if (completion) {
+                    completion(nil,nil,nil,nil);
```
samgreen: Should be `completion(nil, nil, nil, nil);`

jaredmixpanel:
@samgreen Style fixed and yes, I think this would resolve #420 with the notable exception that this fix doesn't allow you to programmatically distinguish between success and failure.

samgreen:
Hey @jaredmixpanel it seems like you can check if an error occurred based on the presence of `nil`. In non error cases it should contain an empty array, in error cases it will be `nil`.

Regardless, this is great, and thanks for the help!

peymano:
I'd like to see a clean API that invokes the callback with an `NSError` object.

PR #472, opened 2016-04-13
Added Carthage/Build to the gitignore list

crenwick:
When this framework is added via carthage with the `--use-submodules` flag, it builds it in the mixpanel-iphone/Carthage/Build directory, marking this submodule as `dirty`. Adding `Carthage/Build` to the gitignore list fixes this :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,4 +23,7 @@ profile
 # AppCode
 .idea
 
+#Carthage
+Carthage/Build
+
 HelloMixpanel/build/


samgreen:
Hmm, I'm not sure how I feel about merging this to master. Personally we check in `Carthage/Build` and ignore `Carthage/Checkouts`. 

Does this change your thoughts at all?

crenwick:
I'm not sure I follow. Does that mean I should be able to find `Carthage/Build` in master without building?

What is the recommended way to use this with `Carthage --use-submodules` without marking it as `dirty` on each build?

crenwick:
Oh, do you mean that is your method on iOS projects that use mixpanel, and not this repo specifically?

samgreen:
Yes, specifically we use it in the Mixpanel iOS app.

crenwick:
Definitely down to explore that option. Curious -- are you able to cut down on carthage build times with this technique? I.E, not spending time building frameworks already in the `Carthage/Build` directory?

On a side not, this PR shouldn't effect your build process. Its adding `Carthage/Checkouts/mixpanel-iphone/Carthage/Build` to the gitignore list, and not effecting the iOS app's `Carthage/Build` directory.

samgreen:
Ah you're totally right @crenwick. Sorry for the confusion. Will merge.

This does help with build times, we're also planning to just archive the builds in to the zips for Carthage moving forward so no one has to build. That should improve the workflow considerably.

crenwick:
:+1: , thanks @samgreen 

How do you currently keep carthage from not wiping its previous builds? Is there a flag I'm missing?

The archive system sounds like it would make a fantastic blog post! 😉

samgreen:
Hey @crenwick, have you investigate the `--no-build` flag? This is what I'm referring to, it's documented [here](http://www.mokacoding.com/blog/carthage-no-build/).

As for the archive system it's documented [here](https://github.com/Carthage/Carthage#archive-prebuilt-frameworks-into-one-zip-file). That part is the responsibility of the framework developer, so you don't have to worry too much, but I thought you might be interested in how it works :smile:

crenwick:
Oh, thats a great post and exactly what I've been looking for. Thanks!

PR #475, opened 2016-04-16
code style

Coeur:


samgreen:
Lots of good style changes in here. Will hopefully have time to give it a proper review next week. 

Thanks!

Mixpanel/MPUIImageToNSDictionaryValueTransformer.m
```diff
@@ -95,9 +95,9 @@ - (id)reverseTransformedValue:(id)value
                     NSError *error;
                     NSData *imageData = [NSData dataWithContentsOfURL:imageUrl options:(NSDataReadingOptions)0 error:&error];
                     if (!error) {
-                        image = [UIImage imageWithData:imageData scale:fminf(1.0, [scale floatValue])];
+                        image = [UIImage imageWithData:imageData scale:fminf(1.0, scale.floatValue)];
                         @synchronized(imageCache) {
-                            [imageCache setValue:image forKey:imageDictionary[@"url"]];
+                            imageCache[imageDictionary[@"url"]] = image;
```
samgreen: I believe this will throw an exception on `nil` while `setValue:forKey:` does not.

Mixpanel/Mixpanel.m
```diff
@@ -911,51 +910,39 @@ - (id)unarchiveFromFile:(NSString *)filePath
 
 - (void)unarchiveEvents
 {
-    self.eventsQueue = (NSMutableArray *)[self unarchiveFromFile:[self eventsFilePath]];
-    if (!self.eventsQueue) {
-        self.eventsQueue = [NSMutableArray array];
-    }
+    self.eventsQueue = (NSMutableArray *)[self unarchiveFromFile:[self eventsFilePath] asClass:[NSMutableArray class]] ?: [NSMutableArray array];
```
samgreen: This actually changes the logic here. We want to create an `NSMutableArray` if we were unable to load one.

Mixpanel/Mixpanel.m
```diff
@@ -888,11 +884,14 @@ - (void)unarchive
     [self unarchiveEventBindings];
 }
 
-- (id)unarchiveFromFile:(NSString *)filePath
+- (id)unarchiveFromFile:(NSString *)filePath asClass:(Class)class
```
samgreen: Could you just modify this to return `[class alloc] init]` of the unarchived data is nil? This would prevent hiding the ternary operator after each load call.

Mixpanel/Mixpanel.m
```diff
@@ -2018,15 +2006,15 @@ - (NSString *)description
 
 - (NSDictionary *)collectAutomaticPeopleProperties
 {
-    NSMutableDictionary *p = [NSMutableDictionary dictionary];
+    NSMutableDictionary *p = [NSMutableDictionary dictionaryWithDictionary:@{@"$ios_version": [UIDevice currentDevice].systemVersion,
+                                                                             @"$ios_lib_version": VERSION,
+                                                                             }];
+    NSDictionary *infoDictionary = [NSBundle mainBundle].infoDictionary;
+    p[@"$ios_app_version"] = infoDictionary[@"CFBundleVersion"];
+    p[@"$ios_app_release"] = infoDictionary[@"CFBundleShortVersionString"];
```
samgreen: Again, these will throw on a `nil` value and cause an exception.

samgreen:
Some minor comments, but I would LOVE to merge this week. Awaiting your comments @Coeur!

Mixpanel/MPUIImageToNSDictionaryValueTransformer.m
Coeur: No, this subscripting syntax for dictionaries does not throw on `nil`: it is not like the subscripting syntax for arrays where it can throw for out of bounds. And `setValue:forKey:` should be avoided as much as possible, as it may have unpredictable behaviour if the key starts with the '@' character.

Mixpanel/Mixpanel.m
Coeur: No, this subscripting syntax for dictionaries does not throw on `nil`: it is not like the subscripting syntax for arrays where it can throw for out of bounds. And `setValue:forKey:` should be avoided as much as possible, as it may have unpredictable behaviour if the key starts with the '@' character.

Mixpanel/Mixpanel.m
Coeur: Yes, it's possible. I'll do it.

Coeur:
I've merged the changes on unarchiving from file.

samgreen:
Thanks!

PR #481, opened 2016-05-02
Image loading fix for NSBundles

samgreen:
This resolves an issue with the new design of in app notifications. The close button "X" was not loading properly in some configurations.


Mixpanel/MPNotification~iphonelandscape.storyboard
```diff
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="10116" systemVersion="15E65" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES">
     <dependencies>
-        <deployment identifier="iOS"/>
```
patedit: Why do you want to move this from the storyboard as well? Even if we load this storyboard programmatically, I think it's worth to keep this on the storyboard.

HelloMixpanel/HelloMixpanel/HelloMixpanel-Prefix.pch
```diff
@@ -1,7 +0,0 @@
-//
```
patedit: Did you try running the tests? I think they were failing because this file was missing.

Mixpanel/MPNotification~iphonelandscape.storyboard
samgreen: I suppose it was removed automatically. I didn't do this intentionally.

HelloMixpanel/HelloMixpanel/HelloMixpanel-Prefix.pch
samgreen: I'll check that now.

samgreen: All tests are passing!

patedit:
👍

PR #489, opened 2016-05-10
Fix track calls from Apple Watch

ryanseams:
Currently the Apple Watch is looking for a ```messageType``` of ```track:``` when sending Mixpanel messages to the iPhone. Since the code for ```MixpanelWatchOS``` always sends the selector as ```track:properties:```, events are not being properly relayed back to the iPhone.

This PR is to look for the selector ```track:properties:``` from the Apple Watch and allow Mixpanel messages to be sent along to the iPhone.

### DIFF
--- a/Mixpanel/Mixpanel+HostWatchOS.m
+++ b/Mixpanel/Mixpanel+HostWatchOS.m
@@ -14,7 +14,7 @@
 - (void)session:(WCSession *)session didReceiveMessage:(NSDictionary<NSString *, id> *)message {
     NSString *messageType = [Mixpanel messageTypeForWatchSessionMessage:message];
     if (messageType) {
-        if (NSStringFromSelector(@selector(track:properties:))) {
+        if ([messageType isEqualToString:NSStringFromSelector(@selector(track:properties:))]) {
             [self track:message[@"event"] properties:message[@"properties"]];
         }
     }
--- a/Mixpanel/Mixpanel+HostWatchOS.m
+++ b/Mixpanel/Mixpanel+HostWatchOS.m
@@ -14,7 +14,7 @@
 - (void)session:(WCSession *)session didReceiveMessage:(NSDictionary<NSString *, id> *)message {
     NSString *messageType = [Mixpanel messageTypeForWatchSessionMessage:message];
     if (messageType) {
-        if ([messageType isEqualToString:@"track:properties:"]) {
+        if (NSStringFromSelector(@selector(track:properties:))) {
             [self track:message[@"event"] properties:message[@"properties"]];
         }
     }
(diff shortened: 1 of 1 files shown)


Mixpanel/Mixpanel+HostWatchOS.m
```diff
@@ -14,7 +14,7 @@ @implementation Mixpanel (HostWatchOS)
 - (void)session:(WCSession *)session didReceiveMessage:(NSDictionary<NSString *, id> *)message {
     NSString *messageType = [Mixpanel messageTypeForWatchSessionMessage:message];
     if (messageType) {
-        if ([messageType isEqualToString:@"track"]) {
+        if ([messageType isEqualToString:@"track:properties:"]) {
```
samgreen: Can we change this to `NSSelectorFromString(@selector(track:properties:))`?

Quuxplusone: Peanut gallery says: you mean `NSStringFromSelector`, right?

@samgreen, ObjC style question: Is `NSSelectorFromString(@selector(...))` our preferred style in general? I don't see any instances of `NSStringFromSelector(@selector(...))` in mixpanel-iphone yet, but I do see (`git grep '@"[a-z][^ "]*:'`) some instances of e.g.

    NSSelectorFromString(@"imageForState:")
    NSSelectorFromString(@"didMoveToWindow")

that seem like they ought to be refactored into `@selector(...)`. Or is there a reason we don't do that in those cases?
I'm not proposing a specific competing style here; just trying to understand the rationale for our current style.

http://corporationunknown.com/blog/2014/03/24/to-nsstringfromselector-or-not/

samgreen: @Quuxplusone Totally meant `NSStringFromSelector` updated. I actually think it would be worthwhile to change many of those selectors as well. 

Although I agree with many points of the article you linked, I think we can all agree the "stringly" typed variants we are using now are not buying us anything.

ryanseams: Updating to NSStringFromSelector(@selector(track:properties:)) now

Mixpanel/Mixpanel+HostWatchOS.m
```diff
@@ -14,7 +14,7 @@ @implementation Mixpanel (HostWatchOS)
 - (void)session:(WCSession *)session didReceiveMessage:(NSDictionary<NSString *, id> *)message {
     NSString *messageType = [Mixpanel messageTypeForWatchSessionMessage:message];
     if (messageType) {
-        if ([messageType isEqualToString:@"track"]) {
+        if (NSStringFromSelector(@selector(track:properties:))) {
```
samgreen: Sorry, should have been more clear, we want 

`[messageType isEqualToString:NSStringFromSelector(@selector(track:properties:))]`

ryanseams: hehe worked this way, but changing for you 😸

Issue #333, opened 2015-09-01
Crash when fetching device model


ryanseams:
Submitting this on behalf of a customer.

Currently their app is logging crashes when it accesses the line at https://github.com/mixpanel/mixpanel-iphone/blob/v2.6.2/Mixpanel/Mixpanel.m#L297 which can be found in the most recent release at https://github.com/mixpanel/mixpanel-iphone/blob/master/Mixpanel/Mixpanel.m#L871. Since this pod is old the issue could have been resolved, but since the code has not changed I wanted to post this here.

Suggested fix by the customer is to handle this as an exception, but I might argue we simply ignore assignment of the property in this instance. Here is their fix:
```
- (NSString *)deviceModel 
{ 
size_t size; 
sysctlbyname("hw.machine", NULL, &size, NULL, 0); 
char answer[size]; 
@try { 
sysctlbyname("hw.machine", answer, &size, NULL, 0); 
NSString *results = @(answer); 
return results; 
} 
@catch (NSException *exception) { 
return @""; 
} 
}
```

Here is the stack trace:
0	
libsystem_platform.dylib 0x0000000197585a80 OSAtomicDequeue + 8 
1	
libsystem_malloc.dylib 0x000000019752d064 nano_malloc + 40 
2	
libsystem_malloc.dylib 0x000000019751d3e4 malloc_zone_malloc + 112 
3	
CoreFoundation 0x0000000184f96e08 _CFRuntimeCreateInstance + 372 
4	
CoreFoundation 0x0000000185083f28 __CFStringCreateImmutableFunnel3 + 2196 
5	
CoreFoundation 0x0000000184f9857c CFStringCreateWithBytes + 40 
6	
Foundation 0x0000000185eca6b8 -[NSPlaceholderString initWithBytes:length:encoding:] + 140
7	
Foundation 0x0000000185eca608 +[NSString stringWithUTF8String:] + 156 
8	
**Redacted** 0x00000001002e9c8c -[Mixpanel deviceModel] (Mixpanel.m:297) 
9	
**Redacted** 0x00000001002f5a94 -[MixpanelPeople collectAutomaticPeopleProperties] (Mixpanel.m:1745) 
10	
**Redacted** 0x00000001002f5904 -[MixpanelPeople initWithMixpanel:] (Mixpanel.m:1730) 
11	
[...]

PR #500, opened 2016-05-31
avoiding unnecessary try-catch

Coeur:
A cleaner fix for #333 than #446.


patedit:
Thanks for the clean up @Coeur !

PR #504, opened 2016-06-07
Programatically disable A/B test designer

samgreen:
Add enableABTestDesigner property to the public API to allow disabling the AB test designer without recompiling and providing the `DISABLE_MIXPANEL_AB_DESIGNER` macro.

### DIFF
--- a/Mixpanel/Mixpanel.m
+++ b/Mixpanel/Mixpanel.m
@@ -50,11 +50,13 @@
 #endif
 {
     NSUInteger _flushInterval;
+    BOOL _enableABTestDesigner;
 }
 
 #if !defined(MIXPANEL_APP_EXTENSION)
 @property (nonatomic, assign) SCNetworkReachabilityRef reachability;
 @property (nonatomic, strong) CTTelephonyNetworkInfo *telephonyInfo;
+@property (nonatomic, strong) UILongPressGestureRecognizer *testDesignerGestureRecognizer;
 #endif
 
 // re-declare internally as readwrite
@@ -86,7 +88,7 @@
 @property (nonatomic, strong) UIViewController *notificationViewController;
 @property (nonatomic, strong) NSMutableSet *shownNotifications;
 
-@property (nonatomic, strong) id abtestDesignerConnection;
+@property (nonatomic, strong) MPABTestDesignerConnection *abtestDesignerConnection;
 @property (nonatomic, strong) NSSet *variants;
 @property (nonatomic, strong) NSSet *eventBindings;
 
@@ -94,6 +96,7 @@
 @property (atomic, copy) NSString *switchboardURL;
 @property (nonatomic) NSTimeInterval networkRequestsAllowedAfterTime;
 @property (nonatomic) NSUInteger networkConsecutiveFailures;
+
 
 @end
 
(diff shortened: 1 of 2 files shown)


Mixpanel/Mixpanel.m
```diff
@@ -95,6 +97,7 @@ @interface Mixpanel () <UIAlertViewDelegate>
 @property (nonatomic) NSTimeInterval networkRequestsAllowedAfterTime;
 @property (nonatomic) NSUInteger networkConsecutiveFailures;
 
+
```
yinfeiru: intentional?

Mixpanel/Mixpanel.m
```diff
@@ -191,6 +194,7 @@ - (instancetype)initWithToken:(NSString *)apiToken launchOptions:(NSDictionary *
 
         self.decideResponseCached = NO;
         self.showSurveyOnActive = YES;
+        self.enableABTestDesigner = YES;
```
yinfeiru: I've been thinking about the default value of this. It's more of a product question which should involve Nihar but I'd suggest to turn this off by default.

1. It saves battery/data usage for the users
2. Our switchboard is receiving a lot of requests which are accidentally made by the users.

Also, it'd be valuable to figure out the radio of A/B users vs SDK users.

Mixpanel/Mixpanel.m
```diff
@@ -50,11 +50,13 @@ @interface Mixpanel () <UIAlertViewDelegate>
 #endif
 {
     NSUInteger _flushInterval;
+    BOOL _enableABTestDesigner;
 }
 
 #if !defined(MIXPANEL_APP_EXTENSION)
 @property (nonatomic, assign) SCNetworkReachabilityRef reachability;
 @property (nonatomic, strong) CTTelephonyNetworkInfo *telephonyInfo;
+@property (nonatomic, strong) UILongPressGestureRecognizer *testDesignerGestureRecognizer;
```
yinfeiru: I might be missing something but what's the purpose of `testDesignerGestureRecognizer`?

samgreen: We create a reference to it so that we can disable it outside of the scope of it's creation. For instance if the user initializes the library, and we start the A/B connection, then they disable A/B designer, we need to also disable the gesture recognizer to prevent reconnects.

Mixpanel/Mixpanel.m
samgreen: Nope

Mixpanel/Mixpanel.m
samgreen: Let's put the product question aside for a moment.

To both of your points, would changing the gesture resolve your concerns?

samgreen: It is enabled by default on Android correct? So this would be parity with that SDK.

yinfeiru: Yes. If we can improve the gesture, it will definitely resolve both problems.
And it is enabled on Android by default. Do you know when are we going to roll out the gesture change?

Mixpanel/Mixpanel.m
yinfeiru: I see, make sense.

Mixpanel/Mixpanel.m
samgreen: Decision has been made to leave this functionality as is (enabled by default) to achieve feature parity with [mixpanel-android](github.com/mixpanel/mixpanel-android)

PR #506, opened 2016-06-13
Refactor persistence

samgreen:
This migrates all file handling and persistence functionality in to an `MPPersistence` helper. This alos adds a new test suite focused exclusively on persistence.

Tests Status:
screen shot 2016-06-13 at 4 26 02 pm

Code Coverage Report:
screen shot 2016-06-13 at 4 26 26 pm

### DIFF
--- a/Mixpanel/Mixpanel.m
+++ b/Mixpanel/Mixpanel.m
@@ -24,6 +24,7 @@
 #import <UIKit/UIDevice.h>
 
 #import "MPResources.h"
+#import "MPPersistence.h"
 #import "MixpanelExceptionHandler.h"
 #import "MPABTestDesignerConnection.h"
 #import "UIView+MPHelpers.h"
@@ -62,6 +63,7 @@
 // re-declare internally as readwrite
 @property (atomic, strong) MixpanelPeople *people;
 @property (atomic, copy) NSString *distinctId;
+@property (nonatomic, strong) MPPersistence *persistence;
 @property (nonatomic, getter=isValidationEnabled) BOOL validationEnabled;
 @property (nonatomic) AutomaticEventMode validationMode;
 @property (nonatomic) NSUInteger validationEventCount;
@@ -156,6 +158,7 @@ static Mixpanel *sharedInstance;
         // Install uncaught exception handlers first
         [[MixpanelExceptionHandler sharedHandler] addMixpanelInstance:self];
         self.telephonyInfo = [[CTTelephonyNetworkInfo alloc] init];
+        self.persistence = [[MPPersistence alloc] initWithToken:apiToken];
 #endif
         
         self.networkRequestsAllowedAfterTime = 0;
@@ -389,10 +392,8 @@ static __unused NSString *MPURLEncode(NSString *s)
                 r[@"$distinct_id"] = distinctId;
                 [self.peopleQueue addObject:r];
             }
+            [self.persistence archivePeopleQueue:self.peopleQueue];
             [self.people.unidentifiedQueue removeAllObjects];
-            [self archivePeople];
-        }
-        if ([Mixpanel inBackground]) {
             [self archiveProperties];
         }
     });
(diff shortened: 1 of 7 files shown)


Mixpanel/Mixpanel.m
```diff
@@ -389,10 +392,8 @@ - (void)identify:(NSString *)distinctId
                 r[@"$distinct_id"] = distinctId;
                 [self.peopleQueue addObject:r];
             }
+            [self.persistence archivePeopleQueue:self.peopleQueue];
             [self.people.unidentifiedQueue removeAllObjects];
-            [self archivePeople];
-        }
-        if ([Mixpanel inBackground]) {
```
yinfeiru: why remove this check?

Mixpanel/Mixpanel.m
```diff
@@ -469,11 +470,8 @@ - (void)track:(NSString *)event properties:(NSDictionary *)properties
             [self.eventsQueue removeObjectAtIndex:0];
         }
         
-        if ([Mixpanel inBackground]) {
```
yinfeiru: why did we do this before?

yinfeiru: I mean, why did we do `[self archiveEvents]` twice here?

Mixpanel/Mixpanel.m
```diff
@@ -856,80 +801,23 @@ - (void)archiveProperties
...
 
 - (void)unarchive
 {
-    [self unarchiveEvents];
-    [self unarchivePeople];
+    self.eventsQueue = [self.persistence unarchiveEventQueue];
+    self.peopleQueue = [self.persistence unarchivePeopleQueue];
+    
     [self unarchiveProperties];
-    [self unarchiveVariants];
```
yinfeiru: what about `unarchiveVariants` and `unarchiveEventBindings`? we don't need them any more?

yinfeiru: Oh..so they are done in the `unarchiveProperties`. but why this change? since we have separate logic to handle things like `unarchiveVariants`, do you think it makes more sense to separate them here as well?

Mixpanel/Mixpanel.m
```diff
@@ -2072,22 +1951,24 @@ - (void)addPeopleRecordToQueueWithAction:(NSString *)action andProperties:(NSDic
...
                 MixpanelDebug(@"%@ queueing people record: %@", self.mixpanel, r);
-                [strongMixpanel.peopleQueue addObject:r];
-                if (strongMixpanel.peopleQueue.count > 500) {
-                    [strongMixpanel.peopleQueue removeObjectAtIndex:0];
+                [peopleQueue addObject:r];
+                if (peopleQueue.count > 500) {
+                    [peopleQueue removeObjectAtIndex:0];
                 }
+                MPPersistence *persistence = strongMixpanel.persistence;
+                [persistence archivePeopleQueue:peopleQueue];
```
yinfeiru: why do we move this logic inside this `if`? I think we want to archive people if `people.identify` isn't called yet.

Mixpanel/Mixpanel.m
samgreen: This was a bug, it meant we would archive twice if we were in the background.

Mixpanel/Mixpanel.m
samgreen: Trying to keep the refactoring to a minimum. There are many functions that call `unarchiveProperties`, I don't want to go change all of those call sites to also call `unarchiveVariants` and `unarchiveEventBindings`

Mixpanel/Mixpanel.m
samgreen: We're archiving the people queue which is only valid if they have called `identify:`, if they haven't yet identified the user then we need to saved the `unidentifiedQueue` which contains events that still need their `distinctID` updated.

PR #515, opened 2016-06-17
Handled warning in Xcode 8 and iOS 10

philippec:
The function SecRandomCopyBytes now requires us to handle its return value.


samgreen:
I believe this was fixed with #507 

I would prefer not to change the actual behavior by failing with an error.

philippec:
You are correct. In my private fork of mixpanel, I simply put (void) in front of SecRandomCopyBytes, which to me is slightly more elegant than #pragmas. But I wanted to do something better here, for you.

That being said, I was lazy in submitting the fix so I will defer to #507.

PR #516, opened 2016-06-17
Fix for In App notifications on iPhone 4S

samgreen:
This just removes the aspect ratio constraint, and increases the tap area for the close button.


yarneo:
There is a slight difference between what is seen in the web composer and what is seen in the device. Attached are screenshots for your convenience for comparison.
screen shot 2016-06-17 at 11 40 03 am
screen shot 2016-06-17 at 11 41 18 am

yarneo:
Smaller images are being fitted to full screen rather than staying in original sizing
screen shot 2016-06-17 at 11 27 48 am
screen shot 2016-06-17 at 11 27 52 am

PR #518, opened 2016-06-18
Fixes all tvOS tests

samgreen:


Mixpanel/MPNetwork.m
```diff
@@ -59,26 +60,38 @@ - (void)flushQueue:(NSMutableArray *)queue endpoint:(NSString *)endpoint {
...
+                if ([response intValue] == 0) {
+                    MixpanelError(@"%@ %@ api rejected some items", self, endpoint);
+                }
+            }
+            
+            dispatch_semaphore_signal(semaphore);
+        }] resume];
         
-        [self updateNetworkActivityIndicator:NO];
+        dispatch_semaphore_wait(semaphore, DISPATCH_TIME_FOREVER);
```
patedit: Why do we need a semaphore here and not handle this inside the completion block?

samgreen: We want to keep the external API synchronous, while removing the, deprecated, `NSURLConnection` API

patedit:
LGTM

PR #535, opened 2016-06-27
Mark nameTag as deprecated, remove from examples and tests.

samgreen:


yas375:
Would you recommend something to use instead?

samgreen:
Hi @yas375, we weren't aware of this still being used anywhere. It's been deprecated (unofficially) since Dec 2014. Can you explain what you're using it for?

If you absolutely require `nameTag`, you can still set a super property with the key `$mp_name_tag`. Is that helpful?

yas375:
Is the `nameTag` value you show in these places or am I wrong?

samgreen:
Those properties are set with the people properties name `$first_name`, `$last_name`, `$name`

cassiegamm:
Hi @yas375,

By default the reserved properties `$name` or `$first_name `would be used to populate these views. For your convenience, a full list of reserved property names can be found here: https://mixpanel.com/help/questions/articles/special-or-reserved-properties

If it's relevant, you could use the [set_once](https://mixpanel.github.io/mixpanel-iphone/Classes/MixpanelPeople/index.html#//apple_ref/occ/instm/MixpanelPeople/setOnce:) method to set the unique user id and ensure that it will never be overwritten. Otherwise, the standard set method should cover your needs.

yas375:
@cassiegamm thanks

PR #537, opened 2016-06-28
Add MPNotification test with malformed image URL from customer issue.

samgreen:
This test exercises a real code path that caused a crash in v2.9.8 for a customer due to an invalid image URL. This test confirms that the move to `NSURLComponents` did actually resolve that issue.


yinfeiru:
lgtm

PR #545, opened 2016-07-14
First pass at better URL encoding

samgreen:
Remove nonsensical URL encoding from MPNotification.

Remove MPNetworkTests.h and move to MPNetworkPrivate.h, fix tests since there was a small refactor, add test that checks full URL is built correctly for decide changes.

### DIFF
--- a/Mixpanel/MPNetwork.h
+++ b/Mixpanel/MPNetwork.h
@@ -20,4 +20,12 @@
 
 - (void)updateNetworkActivityIndicator:(BOOL)enabled;
 
+- (NSURLRequest *)requestForEndpoint:(NSString *)endpoint
+                        byHTTPMethod:(NSString *)method
+                      withQueryItems:(NSArray <NSURLQueryItem *> *)queryItems;
+
+- (NSURLRequest *)requestForEndpoint:(NSString *)endpoint
+                        byHTTPMethod:(NSString *)method
+                             andBody:(NSString *)body;
+
 @end
--- a/Mixpanel/MPNetwork.h
+++ b/Mixpanel/MPNetwork.h
@@ -8,6 +8,12 @@
 
 #import <Foundation/Foundation.h>
 
+typedef NS_ENUM(NSUInteger, MPNetworkEndpoint) {
+    MPNetworkEndpointTrack,
+    MPNetworkEndpointEngage,
+    MPNetworkEndpointDecide
+};
+
 @interface MPNetwork : NSObject
 
 @property (nonatomic) BOOL shouldManageNetworkActivityIndicator;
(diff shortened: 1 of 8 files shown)


Mixpanel/MPNetwork.h
```diff
@@ -20,4 +20,12 @@
 
 - (void)updateNetworkActivityIndicator:(BOOL)enabled;
 
+- (NSURLRequest *)requestForEndpoint:(NSString *)endpoint
```
patedit: If we keep `queryItems` and `andBody` we don't need to pass the http method. Alternatively, we could just keep `byHTTPMethod` and pass an `id` (not ideal, I prefer the first solution).

Mixpanel/MPNetwork.m
```diff
@@ -115,16 +116,62 @@ - (BOOL)handleNetworkResponse:(NSHTTPURLResponse *)response withError:(NSError *
 }
 
 #pragma mark - Helpers
-- (NSURLRequest *)requestForEndpoint:(NSString *)endpoint withBody:(NSString *)body {
-    NSURL *URL = [self.serverURL URLByAppendingPathComponent:endpoint];
-    NSMutableURLRequest *request = [NSMutableURLRequest requestWithURL:URL];
++ (NSArray<NSURLQueryItem *> *)buildDecideQueryForProperties:(NSDictionary *)properties
+                                              withDistinctID:(NSString *)distinctID
+                                                    andToken:(NSString *)token {
+    NSURLQueryItem *itemVersion = [NSURLQueryItem queryItemWithName:@"version" value:@"1"];
```
patedit: Is this a new param? Why do we need it?

Mixpanel/MPNetwork.m
```diff
@@ -115,16 +116,62 @@ - (BOOL)handleNetworkResponse:(NSHTTPURLResponse *)response withError:(NSError *
...
+    NSURLQueryItem *itemLib = [NSURLQueryItem queryItemWithName:@"lib" value:@"iphone"];
```
patedit: `lib` or `mp_lib`?

Mixpanel/Mixpanel.m
```diff
@@ -982,7 +975,9 @@ - (void)applicationDidEnterBackground:(NSNotification *)notification
         dispatch_group_enter(bgGroup);
         NSString *requestData = [MPNetwork encodeArrayForAPI:@[@{@"event": @"Integration", @"properties": @{@"token": @"85053bf24bba75239b16a601d9387e17", @"mp_lib": @"iphone", @"distinct_id": self.apiToken}}]];
         NSString *postBody = [NSString stringWithFormat:@"ip=%d&data=%@", self.useIPAddressForGeoLocation, requestData];
-        NSURLRequest *request = [self.network requestForEndpoint:@"/track/" withBody:postBody];
+        NSURLRequest *request = [self.network requestForEndpoint:@"/track/"
```
patedit: Could we use constants for this and `decide`?

Mixpanel/MPNetwork.h
samgreen: Agree, will change.

Mixpanel/MPNetwork.m
samgreen: This is an old param. Remember this is for *decide*

Mixpanel/MPNetwork.m
samgreen: `lib` we're building for *decide* and not *track*

Mixpanel/Mixpanel.m
samgreen: Yes, will do.

patedit: And engage, sorry

patedit:
LGTM

PR #587, opened 2016-10-12
fixed it so completion block wasn't called twice for errors in decide…

pchien:
…. minor cleanup so completion blocks were called from same place for error and success


patedit:
LGTM!!

PR #589, opened 2016-10-19
Watch changes

pchien:
Primary change is from communicating with the server via the ios app to communicating directly from the watch. The interface is now more similar to the ios/tvos/app extension interface, though like tvos and app extension it also only has limited support (notably no a/b testing, surveys, push notifications - which should come in from the phone and show up on the watch anyway)


Mixpanel/MixpanelPeople.m
```diff
@@ -32,7 +36,13 @@ - (NSString *)description
 
 - (NSDictionary *)collectAutomaticPeopleProperties
 {
-    NSMutableDictionary *p = [NSMutableDictionary dictionaryWithDictionary:@{@"$ios_version": [UIDevice currentDevice].systemVersion,
+    NSMutableDictionary *p = [NSMutableDictionary dictionaryWithDictionary:@{
+                                                                             @"$ios_version":
+#if defined(MIXPANEL_WATCH_EXTENSION)
```
patedit: Seems weird that we have a file dedicated to collect watchOS properties for events but here we don't do it.

Mixpanel/Mixpanel.m
```diff
@@ -838,26 +849,33 @@ - (NSDictionary *)collectAutomaticProperties
                                   @"mp_lib": @"iphone",
                                   @"$lib_version": [self libVersion],
                                   @"$manufacturer": @"Apple",
+#if !defined(MIXPANEL_WATCH_EXTENSION)
```
patedit: i'd rather have only one `defined` here and not pick properties here and there depending on the target. Have you considered modularizing how we collect the properties in a way that is cleaner to know what properties will be included depending on the target?

Mixpanel/Mixpanel.m
```diff
@@ -228,6 +234,9 @@ - (void)identify:(NSString *)distinctId
         }
         [self archiveProperties];
     });
+#if defined(MIXPANEL_WATCH_EXTENSION)
+    [self flush];
```
pchien: since the watch doesn't flush on a timer identify wouldn't be sent until the next track event. maybe that's okay though? if that's okay then we don't need this

Mixpanel/Mixpanel.m
```diff
@@ -838,26 +849,33 @@ - (NSDictionary *)collectAutomaticProperties
...
+    [p addEntriesFromDictionary:[MixpanelWatchProperties collectAutomaticProperties]];
+#endif
     return [p copy];
 }
 
 + (BOOL)inBackground
 {
-#if !defined(MIXPANEL_APP_EXTENSION)
+#if defined(MIXPANEL_WATCH_EXTENSION)
+    return [WKExtension sharedExtension].applicationState == WKApplicationStateBackground;
```
pchien: oh i didn't realize that was 3.0+. looks like it's only used in A/B testing so i'll remove this

HelloMixpanel/HelloMixpanel.xcodeproj/project.pbxproj
```diff
@@ -1226,9 +1473,9 @@
 				CLANG_WARN_SUSPICIOUS_IMPLICIT_CONVERSION = YES;
 				CODE_SIGN_ENTITLEMENTS = HelloMixpanel/HelloMixpanel.entitlements;
 				CODE_SIGN_IDENTITY = "";
-				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "";
+				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
 				DEFINES_MODULE = YES;
-				DEVELOPMENT_TEAM = "";
+				DEVELOPMENT_TEAM = E8FVX7QLET;
```
pchien: this is what it did when i wanted to build HelloMixpanel to the real phone/watch. is it bad to commit it because it's open source?

Mixpanel/Mixpanel.m
```diff
@@ -818,9 +827,11 @@ + (NSString *)libVersion
 - (NSDictionary *)collectAutomaticProperties
 {
     NSMutableDictionary *p = [NSMutableDictionary dictionary];
-    UIDevice *device = [UIDevice currentDevice];
     NSString *deviceModel = [self deviceModel];
+#if !defined(MIXPANEL_WATCH_EXTENSION)
+    UIDevice *device = [UIDevice currentDevice];
```
pchien: i'll rewrite this section to be cleaner. the properties it uses UIDevice for are the same ones we get from MixpanelWatchProperties so didn't make sense to use WKInterfaceDevice here if we were adding the same properties later

Mixpanel/MixpanelPrivate.h
```diff
@@ -58,6 +58,11 @@
 @property (nonatomic, getter=isValidationEnabled) BOOL validationEnabled;
 #endif
 
+#if !defined(MIXPANEL_WATCH_EXTENSION)
```
pchien: not sure if this one is possible. UIBackgroundTaskIdentifier and UIViewController don't exist in watchOS

Mixpanel/Mixpanel.m
patedit: makes sense!

Mixpanel/Mixpanel.m
pchien: refactored in e529156

Mixpanel/Mixpanel.m
pchien: removed in e529156

Mixpanel/MPNetwork.m
```diff
@@ -278,7 +278,7 @@ + (BOOL)parseHTTPFailure:(NSHTTPURLResponse *)response withError:(NSError *)erro
 }
 
 - (void)updateNetworkActivityIndicator:(BOOL)enabled {
-#if !MIXPANEL_LIMITED_SUPPORT
+#if !(defined(MIXPANEL_APP_EXTENSION) || defined(MIXPANEL_TVOS_EXTENSION) || defined(MIXPANEL_WATCH_EXTENSION))
```
patedit: Macro?

Mixpanel/Mixpanel.h
```diff
@@ -7,7 +7,15 @@
     #define MIXPANEL_TVOS_EXTENSION 1
 #endif
 
-#define MIXPANEL_LIMITED_SUPPORT (defined(MIXPANEL_APP_EXTENSION) || defined(MIXPANEL_TVOS_EXTENSION))
+#if TARGET_OS_WATCH
+    #define MIXPANEL_WATCH_EXTENSION 1
+#endif
+
+#define MIXPANEL_NO_EXCEPTION_HANDLING (defined(MIXPANEL_APP_EXTENSION))
```
patedit: Sweet!

Mixpanel/MPNetwork.m
pchien: i thought about it but it wasn't going to be used elsewhere. i did the same thing in a couple other places where there wouldn't be a macro that could reasonably be reused. happy to change it if you want though

PR #598, opened 2016-10-27
use layoutSubviews instead of didMoveToWindow for table cell ab tests…

pchien:
… so they will re-update when cells are reused


patedit:
LGTM, thanks man!

PR #607, opened 2016-11-04
Initialize mixpanel instances with sharedInstanceWithToken

patedit:


Mixpanel/Mixpanel.m
```diff
@@ -25,21 +25,21 @@
...
 #else
-        const NSUInteger flushInterval = 60;
+    const NSUInteger flushInterval = 60;
 #endif
-        
-        sharedInstance = [[super alloc] initWithToken:apiToken launchOptions:launchOptions andFlushInterval:flushInterval];
-    });
-    return sharedInstance;
+
+    return [[super alloc] initWithToken:apiToken launchOptions:launchOptions andFlushInterval:flushInterval];
```
pchien: probably should use `self alloc` instead of `super alloc`

Mixpanel/Mixpanel.m
```diff
@@ -25,21 +25,21 @@
...
 
-static Mixpanel *sharedInstance;
+static NSMutableDictionary *instances;
+
 + (Mixpanel *)sharedInstanceWithToken:(NSString *)apiToken launchOptions:(NSDictionary *)launchOptions
 {
-    static dispatch_once_t onceToken;
-    dispatch_once(&onceToken, ^{
-        
+    if (instances && [instances objectForKey:apiToken]) {
```
pchien: i think `dictionary[key]` is standard instead of `objectForKey`. also you don't need to check if `instances` because if it doesn't exist `objectForKey` will be nil by default

Mixpanel/Mixpanel.m
```diff
@@ -49,10 +49,33 @@ + (Mixpanel *)sharedInstanceWithToken:(NSString *)apiToken
 
 + (Mixpanel *)sharedInstance
 {
-    if (sharedInstance == nil) {
-        MPLogWarning(@"sharedInstance called before sharedInstanceWithToken:");
+    if (!instances || instances.count == 0) {
```
pchien: same here can just check `count == 0` because it will be `nil` if `instances` doesn't exist

Mixpanel/Mixpanel.m
```diff
@@ -49,10 +49,33 @@ + (Mixpanel *)sharedInstanceWithToken:(NSString *)apiToken
...
+        MPLogWarning(@"sharedInstance called before creating a Mixpanel instance");
+        return nil;
+    }
+
+    if (instances.count > 1) {
+        MPLogWarning(@"sharedInstance can only be called when you have one Mixpanel instance");
+        return nil;
+    }
+
+    return [instances objectForKey:[[instances keyEnumerator] nextObject]];
```
pchien: since there's only one i think `[[instances allValues] firstObject]` is clearer and has the same performance

Mixpanel/Mixpanel.m
```diff
@@ -49,10 +49,33 @@ + (Mixpanel *)sharedInstanceWithToken:(NSString *)apiToken
...
+
+- (instancetype)init
+{
+    if (self = [super init]) {
+        self.eventsQueue = [NSMutableArray array];
+        self.peopleQueue = [NSMutableArray array];
+        self.timedEvents = [NSMutableDictionary dictionary];
+        self.shownSurveyCollections = [NSMutableSet set];
+        self.shownNotifications = [NSMutableSet set];
+        if (!instances) {
```
pchien: maybe put this in `dispatch_once` instead? it's effectively as fast as checking nil but also gets thread safety for free (i read an interesting article i could probably find if you want more details on why)

Mixpanel/Mixpanel.m
```diff
@@ -123,6 +142,7 @@ - (instancetype)initWithToken:(NSString *)apiToken launchOptions:(NSDictionary *
             [self trackPushNotification:remoteNotification event:@"$app_open"];
         }
 #endif
+        [instances setObject:self forKey:apiToken];
```
pchien: i prefer bracket notation for dictionaries in general

pchien:
lgtm

Issue #608, opened 2016-11-10
logging is always enabled


masawo:
**Integration Method:** CocoaPods
**Xcode Version:** 8.1
**Library Version:** 3.0.6 
**Platform:** iOS
**Device:** Simulator / iPhone 6s
**Language:** Objective-C

**Description:** 
It seems that logging is enabled, in our every biuld scheme.

I think that there is a cause : https://github.com/mixpanel/mixpanel-iphone/pull/607

**Expected Behavior:** 
logging will turn on with Build Settings to `MIXPANEL_DEBUG=1`

pchien:
@masawo thanks for letting us know. We'll put out a fix with the next release but for now `[[Mixpanel sharedInstance] setEnableLogging:NO]` should turn it off.

PR #609, opened 2016-11-10
turn logging off by default as per documentation

pchien:
should fix https://github.com/mixpanel/mixpanel-iphone/issues/608

### DIFF
--- a/Mixpanel/Mixpanel.m
+++ b/Mixpanel/Mixpanel.m
@@ -96,8 +96,6 @@ static NSString *defaultProjectToken;
         self.telephonyInfo = [[CTTelephonyNetworkInfo alloc] init];
 #endif
 #endif
-        MPSetLoggingEnabled(YES);
-
         self.apiToken = apiToken;
         _flushInterval = flushInterval;
         self.useIPAddressForGeoLocation = YES;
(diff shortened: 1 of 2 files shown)


patedit:
LGTM

Mixpanel/Mixpanel.m
```diff
@@ -96,8 +96,6 @@ - (instancetype)initWithToken:(NSString *)apiToken launchOptions:(NSDictionary *
         self.telephonyInfo = [[CTTelephonyNetworkInfo alloc] init];
 #endif
 #endif
-        MPSetLoggingEnabled(YES);
```
yinfeiru: I don't have the full context but according to the title of the pr `turn logging off by default as per documentation`, don't you want to do `MPSetLoggingEnabled(NO)` instead of removing the call here and the definition? Our users might want to turn it on later.

pchien: this was a duplicate method of `Mixpanel setEnableLogging:`. `enableLogging` promised to be off by default but this turned it on by default. they can still set it through the other method (not sure this one is even exposed)

yinfeiru:
lgtm

PR #621, opened 2016-12-12
Deprecate surveys

pchien:


.travis.yml
```diff
@@ -27,5 +27,5 @@ script:
 
 # Run `pod lib lint` if specified
 - if [ $POD_LINT == "YES" ]; then
-    pod lib lint;
+    pod lib lint --allow-warnings;
```
patedit: we may want to add a comment here so we don't forget to remove this after we remove survey support.

patedit:
we probably want to update the documentation as well

wileykestner:
@pchien @patedit 

This commit is causing my host project to show the mixpanel deprecation warnings even though the host project doesn't use the deprecated APIs.  Is there any way to silence the deprecation warnings from the perspective of the host project?

pchien:
@wileykestner which version are you on? We had issues with this on 3.0.8 right after we deprecated but fixed them shortly after. See https://github.com/mixpanel/mixpanel-iphone/issues/629. If you still have an issue with it, could you create an issue with integration method, version, etc so we can track it better? Thanks!

PR #622, opened 2016-12-15
Adding identify: safety check

jaredmixpanel:
Ported the safety check from our js lib, which only changes the distinct_id when calling identify if the given distinct_id is different from the existing distinct_id and the alias. Required creating an alias property.

For reference: https://github.com/mixpanel/mixpanel-js/blob/master/src/mixpanel-core.js#L1339-L1343

### DIFF
--- a/Mixpanel/Mixpanel.m
+++ b/Mixpanel/Mixpanel.m
@@ -278,8 +278,13 @@ static NSString *defaultProjectToken;
     }
     
     dispatch_async(self.serialQueue, ^{
-        self.distinctId = distinctId;
-        self.people.distinctId = distinctId;
+        // identify only changes the distinct id if it doesn't match either the existing or the alias;
+        // if it's new, blow away the alias as well.
+        if (distinctId != self.distinctId && distinctId != self.alias) {
+            self.alias = nil;
+            self.distinctId = distinctId;
+            self.people.distinctId = distinctId;
+        }
         if (self.people.unidentifiedQueue.count > 0) {
             for (NSMutableDictionary *r in self.people.unidentifiedQueue) {
                 r[@"$distinct_id"] = distinctId;
@@ -305,8 +310,16 @@ static NSString *defaultProjectToken;
         MPLogError(@"%@ create alias called with empty distinct id: %@", self, distinctID);
         return;
     }
-    [self track:@"$create_alias" properties:@{ @"distinct_id": distinctID, @"alias": alias }];
-    [self flush];
+    if (alias != distinctID) {
+        self.alias = alias;
+        [self track:@"$create_alias" properties:@{ @"distinct_id": distinctID, @"alias": alias }];
+        dispatch_async(self.serialQueue, ^{
+            [self archiveProperties];
+        });
+        [self flush];
+    } else {
+        MPLogWarning(@"alias matches distinctID - skipping api call.");
+    }
 }
 
 - (void)track:(NSString *)event
@@ -486,6 +499,7 @@ static NSString *defaultProjectToken;
         self.distinctId = [self defaultDistinctId];
         self.superProperties = [NSMutableDictionary dictionary];
         self.people.distinctId = nil;
+        self.alias = nil;
         self.people.unidentifiedQueue = [NSMutableArray array];
         self.eventsQueue = [NSMutableArray array];
         self.peopleQueue = [NSMutableArray array];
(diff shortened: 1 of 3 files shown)


Mixpanel/Mixpanel.m
```diff
@@ -278,8 +278,13 @@ - (void)identify:(NSString *)distinctId
...
+        // identify only changes the distinct id if it doesn't match either the existing or the alias;
+        // if it's new, blow away the alias as well.
+        if (distinctId != self.distinctId && distinctId != self.alias) {
+            self.alias = nil;
+            self.distinctId = distinctId;
+            self.people.distinctId = distinctId;
+        }
         if (self.people.unidentifiedQueue.count > 0) {
             for (NSMutableDictionary *r in self.people.unidentifiedQueue) {
                 r[@"$distinct_id"] = distinctId;
```
pchien: i haven't looked into this a ton but at first glance it's changing a bunch of `$distinct_id` to something that you might not be setting as your actual distinct id. `unidentifiedQueue` is filled by `addPeopleRecordToQueueWithAction:andProperties:`. do we want this to happen if `distinctId` wasn't changed? (i don't know the right answer)

Mixpanel/Mixpanel.m
```diff
@@ -305,8 +310,16 @@ - (void)createAlias:(NSString *)alias forDistinctID:(NSString *)distinctID
         MPLogError(@"%@ create alias called with empty distinct id: %@", self, distinctID);
         return;
     }
-    [self track:@"$create_alias" properties:@{ @"distinct_id": distinctID, @"alias": alias }];
-    [self flush];
+    if (alias != distinctID) {
```
pchien: use `isEqualToString:` here too

Mixpanel/Mixpanel.m
```diff
@@ -305,8 +310,16 @@ - (void)createAlias:(NSString *)alias forDistinctID:(NSString *)distinctID
...
+        self.alias = alias;
+        [self track:@"$create_alias" properties:@{ @"distinct_id": distinctID, @"alias": alias }];
+        dispatch_async(self.serialQueue, ^{
+            [self archiveProperties];
+        });
+        [self flush];
+    } else {
+        MPLogWarning(@"alias matches distinctID - skipping api call.");
```
pchien: maybe print out the alias/distinctId here for convenience

Mixpanel/Mixpanel.m
```diff
@@ -305,8 +310,16 @@ - (void)createAlias:(NSString *)alias forDistinctID:(NSString *)distinctID
         MPLogError(@"%@ create alias called with empty distinct id: %@", self, distinctID);
         return;
     }
-    [self track:@"$create_alias" properties:@{ @"distinct_id": distinctID, @"alias": alias }];
-    [self flush];
+    if (alias != distinctID) {
+        self.alias = alias;
```
pchien: i think setting this should be in the same `dispatch_async` block as `archiveProperties`. whether that's before or after `track` is up to you, but that way it'll be set and archived and tracked all in sequence instead of being set potentially way before the rest depending on how much is in the queue at the time

Mixpanel/Mixpanel.m
```diff
@@ -278,8 +278,13 @@ - (void)identify:(NSString *)distinctId
     }
     
     dispatch_async(self.serialQueue, ^{
-        self.distinctId = distinctId;
-        self.people.distinctId = distinctId;
+        // identify only changes the distinct id if it doesn't match either the existing or the alias;
+        // if it's new, blow away the alias as well.
+        if (distinctId != self.distinctId && distinctId != self.alias) {
```
jaredmixpanel: Doh... I've let my ObjC get rusty.

Mixpanel/Mixpanel.m
jaredmixpanel: Good catch. We don't want that... should use self.distinctId instead.

Mixpanel/Mixpanel.m
jaredmixpanel: Make sense.

Mixpanel/Mixpanel.m
jaredmixpanel: done.

Mixpanel/Mixpanel.m
```diff
@@ -305,8 +310,16 @@ - (void)createAlias:(NSString *)alias forDistinctID:(NSString *)distinctID
         MPLogError(@"%@ create alias called with empty distinct id: %@", self, distinctID);
         return;
     }
-    [self track:@"$create_alias" properties:@{ @"distinct_id": distinctID, @"alias": alias }];
-    [self flush];
+    if (![alias isEqualToString:distinctID]) {
+        [self track:@"$create_alias" properties:@{ @"distinct_id": distinctID, @"alias": alias }];
+        [self flush];
+        dispatch_async(self.serialQueue, ^{
```
pchien: ah i think you want this above flush and maybe above track. the reason being that flush has a blocking network request in it so your setting/archiving could wait a long time. it's possible above track is best because track potentially flushes inside of it.

jaredmixpanel: Fixed.

pchien:
@jaredmixpanel this looks fine to me now but i'll wait for https://github.com/mixpanel/mixpanel-swift/pull/49 to get reviewed too before merging just to get it double checked

PR #636, opened 2017-01-23
Don't set queryItems if they're nil

gmerinojimenez:
Hello, 

According to the documentation: “Passing nil to setQueryItems removes the query component of the NSURLComponents object.”

If the serverUrl is custom and it already has query parameters, they will be overridden when setting queryItems to nil, that’s the situation in my company, so I guess that it could potentially happen to someone else.

With this change we should not modify current behavior and we’ll allow developers to use query parameters in their serverUrl

Thank you so much for your time.


pchien:
@gmerinojimenez Thanks for the pr!

PR #638, opened 2017-01-31
In-Apps 2.0 implementation

patedit:


HelloMixpanel/HelloMixpanel.xcodeproj/xcshareddata/xcschemes/[iOS] HelloMixpanel.xcscheme
```diff
@@ -97,7 +97,12 @@
       <EnvironmentVariables>
          <EnvironmentVariable
             key = "CFNETWORK_DIAGNOSTICS"
-            value = "1"
+            value = "0"
```
Comment (patedit) [inline HelloMixpanel/HelloMixpanel.xcodeproj/xcshareddata/xcschemes/[iOS] HelloMixpanel.xcscheme]: Did this to avoid too many logs.

Mixpanel/MPNotificationViewController.m
```diff
@@ -215,44 +234,39 @@ - (void)viewDidLoad
...
+            [self.imageView setImage:tintedImage];
+            self.imageView.tintColor = UIColorFromRGB(notification.imageTintColor);
             self.imageView.hidden = NO;
         } else {
             self.imageView.hidden = YES;
         }
 
         if (self.notification.body != nil) {
-            self.bodyLabel.text = self.notification.body;
+            self.bodyLabel.text = notification.body;
```
pchien: should we also change all the checks to `notification.x != nil` since we changed what we're setting from `self.notification.body` to `notification.body`? happens in other parts of this file too

Mixpanel/MPNotificationViewController.m
```diff
@@ -440,6 +449,10 @@ - (instancetype)initWithCoder:(NSCoder *)aDecoder
     if (self = [super initWithCoder:aDecoder]) {
         _maskLayer = [GradientMaskLayer layer];
         [self.layer setMask:_maskLayer];
+        [_maskLayer setColors:@[(id)[UIColor blackColor].CGColor,(id) [UIColor blackColor].CGColor,(id) [UIColor clearColor].CGColor,(id) [UIColor clearColor].CGColor]];
```
pchien: why did you have to cast to id? is that safe to do with CGColor? and if you need an object why not just use the UIColor that you're originally getting the CGColor from?

Mixpanel/MPNotification.m
```diff
@@ -6,158 +6,97 @@
...
+        
+        NSURL *imageURL = nil;
+        NSString *imageURLString = object[@"image_url"];
+        if (imageURLString != nil && ![imageURLString isKindOfClass:[NSNull class]]) {
+            if (![imageURLString isKindOfClass:[NSString class]]) {
+                [MPNotification logNotificationError:@"image url" withValue:imageURLString];
+                return nil;
+            }
+            
+            NSString *escapedURLString = [imageURLString stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet URLQueryAllowedCharacterSet]];
```
patedit: sanity check in case either the url is not encoded on the sql and we don't encode the url on the decide response. if it's encoded this won't do anything :)

Mixpanel/UIColor+MPColor.h
```diff
@@ -1,5 +1,7 @@
 #import <UIKit/UIKit.h>
 
+#define UIColorFromRGB(rgbValue) [UIColor colorWithRed:((float)((rgbValue & 0xFF0000) >> 16))/255.0 green:((float)((rgbValue & 0xFF00) >> 8))/255.0 blue:((float)(rgbValue & 0xFF))/255.0 alpha:((float)((rgbValue & 0xFF000000) >> 24))/255.0];
```
patedit: i was a bit lazy - you're right, let's try to avoid macros :)

Mixpanel/MPNotificationViewController.m
```diff
@@ -94,42 +101,61 @@ - (void)viewDidLoad
...
+
+        if (notification.buttons.count == 2) {
+            [self setUpButtonView:self.secondButton withData:notification.buttons[1] forIndex:1];
+        } else {
+            [[NSLayoutConstraint constraintWithItem:_secondButtonContainer attribute:NSLayoutAttributeWidth relatedBy:NSLayoutRelationEqual toItem:nil attribute:NSLayoutAttributeNotAnAttribute multiplier:1 constant:0] setActive:YES];
         }
+        
+        self.viewMask.backgroundColor = UIColorFromRGB(notification.backgroundColor);
+        self.viewMask.clipsToBounds = YES;
+        self.viewMask.layer.cornerRadius = 6.f;
```
patedit: I personally prefer to specify what type of number a variable is expecting so it's easier for someone else to modify it without having to check for `cornerRadius` type.

pchien: makes sense

Mixpanel/MPNotificationViewController.m
patedit: `setColors` is expecting an array of `CGColorRef` and `CGColor` is an struct :( I know, it looks very lame but wanted to make clear we are passing the right struct. I'll pass `UIcolor` so it does not as ugly

Mixpanel/MPNotificationViewController.m
```diff
@@ -151,33 +150,53 @@ - (void)buttonTapped:(UIButton *)button {
...
+    _window = [[UIWindow alloc] initWithFrame:CGRectMake(0, [UIScreen mainScreen].bounds.size.height, [UIScreen mainScreen].bounds.size.width, [UIScreen mainScreen].bounds.size.height)];
+    _window.windowLevel = UIWindowLevelAlert;
+    _window.rootViewController = self;
+    [_window setHidden:NO];
+
+    CGRect windowFrame = _window.frame;
+    windowFrame.origin.y -= [UIScreen mainScreen].bounds.size.height;
+
+    [UIView animateWithDuration:0.25 animations:^{
+        [self->_window setFrame:windowFrame];
```
pchien: oh whoa. is there a reason we aren't doing this as a property? the `->ivar` notation is pretty non-standard at this point

Mixpanel/MPNotificationViewController.m
```diff
@@ -151,33 +150,53 @@ - (void)buttonTapped:(UIButton *)button {
...
+    CGRect windowFrame = _window.frame;
+    windowFrame.origin.y += [UIScreen mainScreen].bounds.size.height;
+
+    [UIView animateWithDuration:0.5 animations:^{
+        [self->_window setFrame:windowFrame];
+    } completion:^(BOOL finished) {
+        [self->_window setHidden:YES];
+        [self->_window removeFromSuperview];
+        self->_window = nil;
+        if (completion != NULL) {
```
pchien: objc is a special snowflake and uses `nil` instead of `NULL`

Mixpanel/MPNotificationViewController.m
```diff
@@ -151,33 +150,53 @@ - (void)buttonTapped:(UIButton *)button {
...
+    }];
+}
+
+- (BOOL)shouldAutorotate {
     return NO;
 }
 
-- (UIStatusBarStyle)preferredStatusBarStyle
-{
+- (UIStatusBarStyle)preferredStatusBarStyle {
```
pchien: 👍  style consistency

PR #641, opened 2017-02-09
Set config track ip

paulz:
Configures if IP location should be sent to Mixpanel server.


pchien:
@paulz could you make a new pull request with just the changes that you made? I'm not sure what happened but this one looks like it has a bunch of history from 2014 and over 1000 commits. I haven't checked the diff yet but I don't want to merge something that looks like it branched off over 2 years ago.

PR #651, opened 2017-03-01
Suppressing internal deprecated API usage warnings.

wkiefer:
Our team is using mixpanel and these warnings were too noisy for our warning-free environment.

thanks!


pchien:
Hey @wkiefer we'll be removing surveys in March (https://mixpanel.com/help/reference/ios#surveys) so you don't have long to wait before the warnings will be gone forever! In that light though I think we'll just wait to do that instead of merge this PR since we probably won't do many releases before then anyway.

wkiefer:
Sounds good - thanks @pchien

PR #654, opened 2017-03-08
Fix bug when using "Hide Text" option for in-apps

jaredmixpanel:
Currently, checking "Hide Text" will send null for the body and title which isn't an NSString and causes the notification not to be shown.


jaredmixpanel:
fixed in https://github.com/mixpanel/mixpanel-iphone/pull/655

PR #656, opened 2017-03-10
General warning fixes for v3.1.2

hassanain-idris:
Good to hear that you are going to remove the survey deprecated code soon, The reason behind this PR is to eliminate/silence all the warnings from mixpanel code. Since the project I am involved in is treating warnings as errors (and hence those prevents compiling and running). Plus no warnings is a good thing :)

Anyways, I updated this PR to only include general warnings fixes.


pchien:
Thanks for doing this @hassanain-idris! We're planning on removing surveys from our product this month, so I'd rather not change any of that code. If you want to make a different PR with just the other warning fixes, that would be much appreciated!

pchien:
thanks for doing this!

PR #659, opened 2017-03-14
Fixed static analyzer warnings

philippec:
Marked +sharedInstance as nullable since it can return nil.

Properly initialized distinctId to avoid a macOS warning.


philippec:
I also updated the Mac code-signing identity, it was previously "-" which is not a valid identity and causes a build break when running Analyze.

Mixpanel/Mixpanel.m
```diff
@@ -254,9 +254,10 @@ + (void)assertPropertyTypesInArray:(NSArray *)arrayOfProperties depth:(NSUIntege
 
 - (NSString *)defaultDistinctId
 {    
-    NSString *distinctId = [self IFA];
+    NSString *distinctId;
 
 #if !defined(MIXPANEL_WATCHOS) && !defined(MIXPANEL_MACOS)
+    distinctId = [self IFA];
```
pchien: why did you move this?

philippec: If you build for Mac, you would assign a value to distinctId and never use it, which causes a static analyzer warning.

pchien: Ah. But now the case that's outside the `#if #elif #endif` block doesn't have the `[self IFA]` value assigned anymore. If you want to solve this warning, maybe rewrite it to initialize to `[self IFA]` or `[self macOSIdentifier]` initially based on `MIXPANEL_MACOS` and then go through the `#if` and the rest.

philippec: Well, no, that's the warning. You assign it a value, and immediately assign it another value so the first assignment has no effect.

In the olden days I would have initialized it to nil, but that is no longer necessary in the modern runtime IIRC.

philippec: "warning: Value stored to 'distinctId' during its initialization is never read"

pchien: Yes but right now the lines after `#endif` that check `if (!distinctId)` have different behavior because you no longer have it set to `[self IFA]` anymore if `MIXPANEL_WATCHOS` is defined. You could change it to initialize to `[self macOSIdentifier]` if it's macos and `[self IFA]` otherwise instead of inside the `!defined(MIXPANEL_WATCHOS)` block so that it will have the same behavior as before.

philippec: Ah! Of course you are correct. Sorry I did not see this earlier. I have modified the commit to explicitly initialize distinctId for macOS and non-macOS.

philippec:
This version of the PR should cover all cases for initialization of distinctId. Thank you!

pchien:
Thanks!

PR #668, opened 2017-04-29
Changes assert to NSAssert

c1-jlawson:
Updates `assert()` to `NSAssert` as the rest of the assertion calls in the file.

This change prevents users from encountering the `Implicit declaration of function 'assert' is invalid in C99` on line 738 when compiling on some projects.


c1-jlawson:
Made a silly mistake - will reopen when I fix it.

PR #685, opened 2017-06-14
Fix warnings when building in Xcode 9

anderscarling:
This is basically a first stab at the strict prototypes related warnings
that where added in Xcode 9.

I've basically only confirmed that this compiles at this point - so it
should probably be tested to make sure no runtime errors are introduced
etc.

### DIFF
--- a/Mixpanel/MPABTestDesignerConnection.h
+++ b/Mixpanel/MPABTestDesignerConnection.h
@@ -14,7 +14,7 @@ extern NSString *const kSessionVariantKey;
 @property (nonatomic, assign) BOOL sessionEnded;
 
 - (instancetype)initWithURL:(NSURL *)url;
-- (instancetype)initWithURL:(NSURL *)url keepTrying:(BOOL)keepTrying connectCallback:(void (^)())connectCallback disconnectCallback:(void (^)())disconnectCallback;
+- (instancetype)initWithURL:(NSURL *)url keepTrying:(BOOL)keepTrying connectCallback:(void (^)(void))connectCallback disconnectCallback:(void (^)(void))disconnectCallback;
 
 - (void)setSessionObject:(id)object forKey:(NSString *)key;
 - (id)sessionObjectForKey:(NSString *)key;
--- a/Mixpanel/MPABTestDesignerConnection.m
+++ b/Mixpanel/MPABTestDesignerConnection.m
@@ -43,11 +43,11 @@ static NSString * const kFinishLoadingAnimationKey = @"MPConnectivityBarFinishLo
     NSOperationQueue *_commandQueue;
     UIView *_recordingView;
     CALayer *_indeterminateLayer;
-    void (^_connectCallback)();
-    void (^_disconnectCallback)();
+    void (^_connectCallback)(void);
+    void (^_disconnectCallback)(void);
 }
 
-- (instancetype)initWithURL:(NSURL *)url keepTrying:(BOOL)keepTrying connectCallback:(void (^)())connectCallback disconnectCallback:(void (^)())disconnectCallback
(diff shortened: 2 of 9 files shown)


pchien:
Thanks @anderscarling! Will take a look

pchien:
@anderscarling I didn't see all the same errors as you when using the demo project. I only saw errors in Mixpanel.m/h because the other ones aren't public header files. I'll make a different PR with just those changes - didn't want to make the other ones because I'm less sure about those changes and also I didn't find a way to define a generic block (for `swizzleBlock`) and wasn't sure about casting back and forth there.

pchien:
@anderscarling just released with the changes I made. It got rid of all the errors for the demo project, but definitely let me know if you see more. I'm still not sure why you got errors for the private files.

anderscarling:
@pchien

After some research it seems like the warnings stems from the fact that I used the module-import syntax `@import Mixpanel` and `@import Mixpanel.MPTweakInline`. Using traditional #include's causes no warnings in my project.

This is the compiler output I get when I build my project the `@import` syntax:

```
While building module 'Mixpanel' imported from /path/to/my/class.m:25:
In file included from <module-includes>:1:
In file included from /path/to/DerivedData/AppName-randomtext/Build/Products/Debug-iphoneos/Mixpanel/Mixpanel.framework/Headers/Mixpanel-umbrella.h:24:
/path/to/DerivedData/AppName-bxtdburpriswweedaropgvogvvrx/Build/Products/Debug-iphoneos/Mixpanel/Mixpanel.framework/Headers/MPABTestDesignerConnection.h:17:95: warning: this function declaration is not a prototype [-Wstrict-prototypes]
- (instancetype)initWithURL:(NSURL *)url keepTrying:(BOOL)keepTrying connectCallback:(void (^)())connectCallback disconnectCallback:(void (^)())disconnectCallback;
                                                                                              ^
                                                                                               void
/path/to/DerivedData/AppName-randomtext/Build/Products/Debug-iphoneos/Mixpanel/Mixpanel.framework/Headers/MPABTestDesignerConnection.h:17:142: warning: this function declaration is not a prototype [-Wstrict-prototypes]
- (instancetype)initWithURL:(NSURL *)url keepTrying:(BOOL)keepTrying connectCallback:(void (^)())connectCallback disconnectCallback:(void (^)())disconnectCallback;
                                                                                                                                             ^
                                                                                                                                              void
While building module 'Mixpanel' imported from /path/to/my/class.m:25:
In file included from <module-includes>:1:
In file included from /path/to/DerivedData/AppName-randomtext/Build/Products/Debug-iphoneos/Mixpanel/Mixpanel.framework/Headers/Mixpanel-umbrella.h:55:
/path/to/DerivedData/AppName-randomtext/Build/Products/Debug-iphoneos/Mixpanel/Mixpanel.framework/Headers/MPSwizzler.h:14:29: warning: this function declaration is not a prototype [-Wstrict-prototypes]
typedef void (^swizzleBlock)();
                            ^
                             void
3 warnings generated.
```

anderscarling:
I'm thinking  this might be because your pod spec does neither explicitly define a [module_map](https://guides.cocoapods.org/syntax/podspec.html#module_map) nor assigns [public_headers](https://guides.cocoapods.org/syntax/podspec.html#public_header_files).

This would, according to the combined documentation in those two links, cause a module map to be created with all headers in source_files (i.e. `s.ios.source_files  = 'Mixpanel/**/*.{m,h}'`).

That sounds to me like the reason all private headers are imported.

anderscarling:
@pchien Any comment on this?

pchien:
I haven't had time to look into it yet but I checked and you're right that we aren't setting `module_map` and Cocoapods will treat everything as public then. I'll look into it with the rest of Xcode 9 and iOS 11 soon.

anderscarling:
I think this is all fixed after #698 is merged, so i'll close this pull req.

PR #699, opened 2017-06-29
Specify SKIP_INSTALL = YES on project level

filblue:
SKIP_INSTALL option was set to YES explicitly for every target except target `Mixpanel`, which caused issues with archiving client app. This sets SKIP_INSTALL = YES on project level, propagating to all targets.


pchien:
Makes sense. Thanks for doing this @filblue.

PR #708, opened 2017-07-25
Allow user to explicitly set people distinct_id

wgins:


Mixpanel/Mixpanel.m
```diff
@@ -300,7 +300,13 @@ - (NSString *)defaultDistinctId
...
 }
 
-- (void)identify:(NSString *)distinctId
+
+- (void)identify:(NSString *)distinctId;
+{
+    [self identify:distinctId usePeople:YES];
+}
+
+- (void)identify:(NSString *)distinctId usePeople(BOOL)usePeopleDistinctId;
```
pchien: oh i didn't notice this before. usePeople needs a colon after it

Mixpanel/Mixpanel.m
```diff
@@ -315,17 +321,21 @@ - (void)identify:(NSString *)distinctId
...
                 self.distinctId = distinctId;
             }
-            self.people.distinctId = distinctId;
-        }
-        if (self.people.unidentifiedQueue.count > 0) {
-            for (NSMutableDictionary *r in self.people.unidentifiedQueue) {
-                r[@"$distinct_id"] = self.distinctId;
-                @synchronized (self) {
-                    [self.peopleQueue addObject:r];
+            if(usePeopleDistinctId) {
```
cmsepul: why is this inside the `if` condition starting at line 319?
Shouldn't the queued updates for `people` be processed even if the provided distinctId is equal to the stored alias @pchien ?

PR #714, opened 2017-08-21
properties for $campaign_open for the cta url and if the button was p…

pchien:
…rimary or secondary. also now always track $campaign_open if they tap on either button or mini, instead of only if there was a ctal url

### DIFF
--- a/Mixpanel/MPNotificationViewController.m
+++ b/Mixpanel/MPNotificationViewController.m
@@ -163,7 +163,15 @@
 }
 
 - (void)buttonTapped:(UIButton *)button {
-    [self.delegate notificationController:self wasDismissedWithCtaUrl:((MPTakeoverNotification *)self.notification).buttons[button.tag].ctaUrl];
+    MPTakeoverNotification *takeoverNotification = (MPTakeoverNotification *)self.notification;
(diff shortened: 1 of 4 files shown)


Mixpanel/MPNotificationViewController.m
```diff
@@ -163,7 +163,15 @@ - (void)setUpButtonView:(UIButton *)buttonView withData:(MPNotificationButton *)
 }
 
 - (void)buttonTapped:(UIButton *)button {
-    [self.delegate notificationController:self wasDismissedWithCtaUrl:((MPTakeoverNotification *)self.notification).buttons[button.tag].ctaUrl];
+    MPTakeoverNotification *takeoverNotification = (MPTakeoverNotification *)self.notification;
+    NSString *whichButton = @"primary";
+    if (takeoverNotification.buttons.count == 2) {
+        whichButton = button.tag == 0 ? @"secondary" : @"primary";
```
patedit: just double checking that in android `buttonIndex=0` is `primary` and in iOS seems the other way around

pchien: apparently it's right here and wrong on android

Issue #704, opened 2017-07-13
Any way to disable Automatic Crash Reporting


saurabha5:
**Integration Method:** CocoaPods
**Xcode Version:** 8.3.3
**Library Version:** 3.1.8
**Platform:** iOS 
**Language:** Objective-C
**Description:** Disable Automatic Crash Reporting 
**Expected Behavior:** Mixpanel is having conflict with Apteligent/Crittercism, so is there anyway to disable Crash Reporting from Mixpanel ?

pchien:
@saurabha5 what's the conflict you're getting? We don't have any flags for disabling crash reporting right now. Could you also try switching the order of initialization between Mixpanel and Apteligent/Crittercism? This shouldn't cause an issue, but Mixpanel's exception handler will call the previously set one when it happens, and potentially Apteligent/Crittercism doesn't do that. At worst I think all that could happen is Mixpanel's isn't called and the other one is though. Let me know what you find out!

saurabha5:
Basically, I don't see crash reported to Apteligent/Crittercism. The troubleshooting steps show that there is a change in NSGetUncaughtExceptionHandler.  Used below to print handler after both Mixpanel and Apteligent is initialized and at various point in the application and this shows the value is changed.

NSLog(@"Handler : %p", NSGetUncaughtExceptionHandler());

As you suggested and I tried before also to switch the order after this Apteligent started reporting the Crashes. Mixpanel first and then Apteligent. If at all there could be a way to completely tell Mixpanel not to register for Exception then it would have been a cleaner solution.

pchien:
It's strange that it was an issue. From https://github.com/mixpanel/mixpanel-iphone/blob/master/Mixpanel/MixpanelExceptionHandler.m#L73 and https://github.com/mixpanel/mixpanel-iphone/blob/master/Mixpanel/MixpanelExceptionHandler.m#L174 it looks like we should be calling the previously set exception handler within our exception handler. So it's expected that the value of the handler is changed, but we shouldn't be cutting Apteligent out of the loop. I'll try to take a look when I get time, but it's good that this helped fix your problem.

banaslee:
We had the same issue, it was stealing most of our crashes from HockeyApp. 
Since HockeyApp was our only crash tracking tool and Mixpanel was just used as an app events tracker, we were setting up HockeyApp first.

[...]

PR #717, opened 2017-09-07
don't unregister exception and signal handlers after any exception/si…

pchien:
…gnal is caught. save previous signal handlers and call them

we weren't calling previous signal handlers before. i'm hoping this fixes the root problem for https://github.com/mixpanel/mixpanel-iphone/issues/704


pchien:
oh wait this still isn't working with crashlytics for signals. closing until i resolve that

PR #737, opened 2017-10-26
Remove Mixpanel-AppExtension from README

emes:
Just a little docfix from 52634ae


pchien:
Oops, thanks for the fix!

Issue #734, opened 2017-10-24
Crash in +[_MPTweakInlineLoader load] with Address Sanitizer turned on


Reggian:
**Integration Method:** CocoaPods
**Xcode Version:**  Version 9.0.1 (9A1004)
**Library Version:** 3.2.0
**Platform:** iOS
**Language:** Objective-C
**Description:**
Crash in `+[_MPTweakInlineLoader load]` on line 133 (`NSString *name = [NSString stringWithString:*entry->name];`)
`*entry->name` is `nil`

When run on iOS Simulator with Address Sanitizer turned on.

rlasante:
yeah just ran into this also, need to make sure that `entry->name != nil` before you try to dereference it

pchien:
Hey @Reggian @rlasante do you guys know why it's storing the tweak with nil name? Did you use a variable for the name in either `MPTweakValue(name_, ...)` or `MPTweakBind(object_, property_, name_, ...)`? I don't know the code behind this super well but it looks like it stores the entries at compile time with the name, so as long as you use string constants for the names it shouldn't end up being `nil` when loading at runtime.

pchien:
@szigetics let's use this thread. And same question from the last post for you in case you have more info. Thanks.

Reggian:
Hi, @pchien, I use it like this:
```
NSString *backgroundColorTweakString = MPTweakValue(@"App Message back color", @"blue");
```

pchien:
That's strange that it would cause an issue, but I'll merge the PR as a safeguard anyway.

pchien:
The PR has been released in the latest version.

PR #738, opened 2017-10-27
Fix Dereference Crash Running in AddressSanitizer

rlasante:
The entry is full of nil objects when running in AddressSanitizer. Don't try to derefence a nil pointer.


rlasante:
resolves https://github.com/mixpanel/mixpanel-iphone/issues/734

PR #756, opened 2017-12-11
only attempt to convert and add push token if present

wgins:
cc: @amitayfeder1

Right now `addPushDeviceToken` blindly attempts to call `pushDeviceTokenToString` which returns `nil` if deviceToken isn't present. Passing `nil` to NSDictionary causes a crash. This change checks if there's actually a push token before attempting to convert to a string. 

We used to handle this scenario in 2.9.9 via the logic here: https://github.com/mixpanel/mixpanel-iphone/blob/v2.9.9/Mixpanel/Mixpanel.m#L2093-L2095

### DIFF
--- a/Mixpanel/MixpanelPeople.m
+++ b/Mixpanel/MixpanelPeople.m
@@ -150,8 +150,9 @@
 
 - (void)addPushDeviceToken:(NSData *)deviceToken
 {
-    if (deviceToken){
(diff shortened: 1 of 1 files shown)


pchien:
@wgins let's check the response of `pushDeviceTokenToString` instead of the token itself. In case something else goes wrong.

PR #757, opened 2017-12-11
urban airship integration support

pchien:


cerupcat:
@patedit Do we need to add any configuration to get this working? I just integrated UA via cocoapods, but `setUrbanAirshipPeopleProp` is never called from `setupIntegrations` since this is never true `if ([integrations containsObject:@"urbanairship"]) {`

patedit:
@cerupcat Thanks for writing in! Yes, you need to connect Urban Airship with Mixpanel. In mixpanel.com, click on the upper right gear button and go into Integrations. Once there, you should enter your Urban Airship credentials in order to enable the integration (more info here: https://help.mixpanel.com/hc/en-us/articles/115005700186)

Once it's connected, you could export your cohorts into UA by following the steps described here: https://help.mixpanel.com/hc/en-us/articles/115005712363

Let me know if you have any other questions!

cerupcat:
Thanks, I had it connected on the wrong project. This makes sense. Thanks for the quick response!

PR #765, opened 2018-02-01
Fix warning (-Wexpansion-to-defined) in Xcode 9.3

anderscarling:
Xcode 9.3 seems to run with -Wexpansion-to-defined by default, which creates a warning in both the Mixpanel target and (due to this happening in headers) in my app target.

I believe this should fix is, but probably if someone more experience in macro expansion trickery could weight in.

Fix is adapted from https://stackoverflow.com/a/42074134


wileykestner:
@anderscarling Have you received any feedback about this PR?  I'm encountering the same issue - have you been using this patch successfully?

anderscarling:
@wileykestner I've heard nothing - and I've not really used the patch in production builds or anything :/

wileykestner:
@anderscarling Thanks for the update - I think we may give it a try in our staging environment.

anderscarling:
Hope it works for you! It feels like simple enough a change not to mess anything up - but as I've no real knowledge in the intricate behaviors of C macros I don't want to make any promises. :)

kdawgwilk:
I was just about to open a PR with a fix for this.  I think it can be simplified down to this:
```objc
#if defined(MIXPANEL_TVOS) || defined(MIXPANEL_WATCHOS) || defined(MIXPANEL_MACOS)
    #define MIXPANEL_NO_REACHABILITY_SUPPORT 1
    #define MIXPANEL_NO_AUTOMATIC_EVENTS_SUPPORT 1
    #define MIXPANEL_NO_NOTIFICATION_AB_TEST_SUPPORT 1
    #define MIXPANEL_NO_CONNECT_INTEGRATION_SUPPORT 1
#elseif defined(MIXPANEL_WATCHOS) || defined(MIXPANEL_MACOS)
    #define MIXPANEL_NO_UIAPPLICATION_ACCESS 1
#elseif defined(MIXPANEL_WATCHOS)
    #define MIXPANEL_FLUSH_IMMEDIATELY 1
#endif
```
You don't need to actually define them as zero since not defining them at all has the same effect

kdawgwilk:
And
```objc
#if defined(MIXPANEL_TVOS) || defined(MIXPANEL_WATCHOS) || defined(MIXPANEL_MACOS)
    #define MIXPANEL_NO_NETWORK_ACTIVITY_INDICATOR 1
#endif
```

kdawgwilk:
Now that Xcode 9.3 is released it seems that this should get released soon so we can rid our projects of these warnings

charliebartel:
What is the status on getting this released?

garnett:
@zihejia could you or other maintainers make a release with this change, please? Xcode 9.3 is now live

anderscarling:
I pushed new commits where i rebased on latest master, and removed the `#else`-cases that `#define`d zero values (as per suggestion from @kdawgwilk).

I didn't pull in the `#elseif`s from @kdawgwilk though, as I think they would actually block `MIXPANEL_NO_UIAPPLICATION_ACCESS` and `MIXPANEL_FLUSH_IMMEDIATELY` from ever being defined.

Travis CI seems to fail for unrelated reason.

zihejia:
@garnett yes, we will be working on this.

PR #768, opened 2018-03-06
put constant MPNotificationTypeTakeover and MPNotificationTypeMini to…

zihejia:
put constant MPNotificationTypeTakeover and MPNotificationTypeMini to public header(Mixpanel.h)


pchien:
why did you create `MPNotificationTypeIsMini` and `MPNotificationTypeIsTakeover` instead of continuing to use `MPNotificationTypeTakeover` and `MPNotificationTypeMini` and just exposing those?

PR #769, opened 2018-03-19
Gdpr support

zihejia:
Background of this change:
In April, 2016, the European Union’s governing bodies passed the General Data Protection Regulation (GDPR), a set of standards and requirements for protecting the privacy and personal information of EU residents. These requirements serve to establish a uniform standard across all EU member states, but do raise the bar above current general practice. They go into effect, in May, 2018.

### DIFF
--- a/HelloMixpanel/HelloMixpanelTests/MixpanelBaseTests.m
+++ b/HelloMixpanel/HelloMixpanelTests/MixpanelBaseTests.m
@@ -37,6 +37,7 @@
     self.mixpanel = [[Mixpanel alloc] initWithToken:kTestToken
                                       launchOptions:nil
                                    andFlushInterval:60];
+    [self deleteOptOutSettingsWithMixpanelInstance:self.mixpanel];
 }
 
 - (void)tearDownMixpanel {
@@ -45,8 +46,16 @@
     stubEngage();
     [self.mixpanel reset];
     [self waitForMixpanelQueues];
+    [self deleteOptOutSettingsWithMixpanelInstance:self.mixpanel];
 
     self.mixpanel = nil;
+}
+
+- (void)deleteOptOutSettingsWithMixpanelInstance:(Mixpanel *)MixpanelInstance {
+    NSFileManager *manager = [NSFileManager defaultManager];
+    NSError *error = nil;
+    NSString *filename = [MixpanelInstance optOutFilePath];
+    [manager removeItemAtPath:filename error:&error];
 }
 
 #pragma mark - Mixpanel Delegate
--- a/HelloMixpanel/HelloMixpanelTests/MixpanelBaseTests.m
+++ b/HelloMixpanel/HelloMixpanelTests/MixpanelBaseTests.m
@@ -37,7 +37,6 @@
     self.mixpanel = [[Mixpanel alloc] initWithToken:kTestToken
                                       launchOptions:nil
                                    andFlushInterval:60];
-    [self deleteOptOutSettingsWithMixpanelInstance:self.mixpanel];
 }
 
 - (void)tearDownMixpanel {
--- a/HelloMixpanel/HelloMixpanelTests/MixpanelOptOutTests.m
+++ b/HelloMixpanel/HelloMixpanelTests/MixpanelOptOutTests.m
@@ -80,7 +80,6 @@
 
 - (void)testAutoTrackShouldBeTriggeredDuringInitializedWithOptedOutNO
 {
-    __block NSInteger trackCount = 0;
     NSDictionary *launchOptions = @{ UIApplicationLaunchOptionsRemoteNotificationKey: @{
                                              @"mp": @{
                                                      @"m": @"the_message_id",
@@ -97,10 +96,6 @@
     
     NSDictionary *p = e[@"properties"];
     XCTAssertEqualObjects(p[@"campaign_id"], @"the_campaign_id", @"campaign_id not equal");
-    
-    XCTAssertTrue(trackCount == 0, @"When opted out, no track call should be ever triggered during initialization.");
-    
-    [MPSwizzler unswizzleSelector:@selector(track:) onClass:[Mixpanel class] named:@"Swizzle Mixpanel.track"];
 }
 
 - (void)testOptInWillAddOptInEvent
@@ -108,10 +103,12 @@
     [self.mixpanel optInTracking];
     XCTAssertFalse([self.mixpanel hasOptedOutTracking], @"The current user should have opted in tracking");
     [self waitForMixpanelQueues];
-    XCTAssertTrue([self.mixpanel.eventsQueue count] == 1, @"When opted in, event queue should have one even(opt in) being queued");
-    if ([self.mixpanel.eventsQueue count]) {
+    if ([self.mixpanel.eventsQueue count] == 1) {
         NSDictionary *event = self.mixpanel.eventsQueue[0];
         XCTAssertEqualObjects(event[@"event"], @"$opt_in", @"When opted in, a track '$opt_in' should have been queued");
+    }
+    else {
+        XCTAssertTrue([self.mixpanel.eventsQueue count] == 1, @"When opted in, event queue should have one even(opt in) being queued");
     }
 }
 
@@ -121,10 +118,12 @@
     XCTAssertFalse([self.mixpanel hasOptedOutTracking], @"The current user should have opted in tracking");
     
     [self waitForMixpanelQueues];
-    XCTAssertTrue([self.mixpanel.eventsQueue count] == 1, @"When opted in, event queue should have one even(opt in) being queued");
-    if ([self.mixpanel.eventsQueue count]) {
+    if ([self.mixpanel.eventsQueue count] == 1) {
         NSDictionary *event = self.mixpanel.eventsQueue[0];
         XCTAssertEqualObjects(event[@"event"], @"$opt_in", @"When opted in, a track '$opt_in' should have been queued");
+    }
+    else {
+        XCTAssertTrue([self.mixpanel.eventsQueue count] == 1, @"When opted in, event queue should have one even(opt in) being queued");
     }
     
     XCTAssertEqualObjects(self.mixpanel.distinctId, @"testDistinctId", @"mixpanel identify failed to set distinct id");
@@ -149,17 +148,17 @@
     XCTAssertEqualObjects(props[@"date"], now);
     XCTAssertEqualObjects(props[@"$app_version"], @"override", @"reserved property override failed");
     
-    XCTAssertTrue([self.mixpanel.eventsQueue count] == 1, @"When opted in, event queue should have one even(opt in) being queued");
-    if ([self.mixpanel.eventsQueue count]) {
+    if ([self.mixpanel.eventsQueue count] == 1) {
         NSDictionary *event = self.mixpanel.eventsQueue[0];
         XCTAssertEqualObjects(event[@"event"], @"$opt_in", @"When opted in, a track '$opt_in' should have been queued");
+    }
+    else {
+        XCTAssertTrue([self.mixpanel.eventsQueue count] == 1, @"When opted in, event queue should have one even(opt in) being queued");
     }
     
     XCTAssertEqualObjects(self.mixpanel.distinctId, @"testDistinctId", @"mixpanel identify failed to set distinct id");
     XCTAssertEqualObjects(self.mixpanel.people.distinctId, @"testDistinctId", @"mixpanel identify failed to set people distinct id");
     XCTAssertTrue(self.mixpanel.people.unidentifiedQueue.count == 0, @"identify: should move records from unidentified queue");
-    
-    
 }
 
 - (void)testHasOptOutTrackingFlagBeingSetProperlyForMultipleInstances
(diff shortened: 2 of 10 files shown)


Mixpanel/Mixpanel.m
```diff
@@ -1144,6 +1173,25 @@ - (NSString *)currentRadio
     return @"";
 }
 
+- (void)optOutTracking
+{
+    [[NSUserDefaults standardUserDefaults] setBool:YES forKey:kMPOptOutFlag];
```
patedit: we should keep track of all distinct_ids that locally have been flagged as opt-out/in

Mixpanel/Mixpanel.m
```diff
@@ -1144,6 +1173,25 @@ - (NSString *)currentRadio
...
+    self.eventsQueue = [NSMutableArray array];
+    self.peopleQueue = [NSMutableArray array];
+    [self archiveEvents];
+    [self archivePeople];
```
pchien: i think anything that's currently on the dispatch queues will still be added to `eventsQueue` and `peopleQueue` and archived. though they won't be sent because of the change you made to `MPNetwork`, if the user opts back in to tracking, those events will be sent, which i don't think we want

Mixpanel/Mixpanel.m
pchien: should the flag be reset on `reset`?

patedit: negative! :)

HelloMixpanel/HelloMixpanelTests/MixpanelOptOutTests.m
```diff
@@ -46,7 +51,7 @@ - (void)testOptOutTrackingWillNotGenerateEventQueue
 {
     stubTrack();
     [self.mixpanel optOutTracking];
-    
+    [self waitForMixpanelQueues];
```
pchien: i think it's actually better to not wait for the queues in this and the following tests that are for events/people stuff. so that it'll actually also test that it works just by calling opt out and then tracking after the way users would do it

HelloMixpanel/HelloMixpanelTests/MixpanelOptOutTests.m
```diff
@@ -70,6 +75,7 @@ - (void)testOptOutTrackingWillSkipIdentify
 {
     stubEngage();
     [self.mixpanel optOutTracking];
+    [self waitForMixpanelQueues];
```
pchien: same for the identify/alias tests

Mixpanel/Mixpanel.m
```diff
@@ -1175,21 +1218,33 @@ - (NSString *)currentRadio
...
 {
-    [[NSUserDefaults standardUserDefaults] setBool:YES forKey:kMPOptOutFlag];
-    self.eventsQueue = [NSMutableArray array];
-    self.peopleQueue = [NSMutableArray array];
-    [self archiveEvents];
-    [self archivePeople];
+    [self.optOutStatus setObject:@YES forKey:self.distinctId];
+    [self archiveOptOut];
+    dispatch_async(self.serialQueue, ^{
+        dispatch_sync(self.networkQueue, ^{ return; });
```
pchien: why does this wait for the network queue?

Mixpanel/Mixpanel.m
```diff
@@ -1175,21 +1218,33 @@ - (NSString *)currentRadio
...
-    [[NSUserDefaults standardUserDefaults] setBool:NO forKey:kMPOptOutFlag];
+    [self.optOutStatus setObject:@NO forKey:self.distinctId];
+    [self archiveOptOut];
 }
 
 - (BOOL)hasOptedOutTracking
 {
-    return [[NSUserDefaults standardUserDefaults] boolForKey:kMPOptOutFlag];
+    NSNumber *optOutStatus = [self.optOutStatus objectForKey:self.distinctId];
+    return optOutStatus? [optOutStatus boolValue]: NO;
```
pchien: i'm pretty sure this is identical to `[optOutStatus boolValue]`

HelloMixpanel/HelloMixpanelTests/MixpanelOptOutTests.m
```diff
@@ -31,6 +31,24 @@ - (void)tearDown
...
+- (void)testHasOptOutTrackingFlagBeingSetProperlyAfterInitializedWithOptedOutYES
+{
+    Mixpanel *mixpanel = [Mixpanel sharedInstanceWithToken:@"abc123Random" optOutTracking:YES];
+    XCTAssertTrue([mixpanel hasOptedOutTracking], @"When optOutTracking is called, the current user should have opted out tracking");
+}
+
+- (void)testHasOptOutTrackingFlagBeingSetProperlyAfterInitializedWithOptedOutNO
+{
+    Mixpanel *mixpanel = [Mixpanel sharedInstanceWithToken:@"abc123Radd" optOutTracking:NO];
+    XCTAssertFalse([mixpanel hasOptedOutTracking], @"When optOutTracking is called, the current user should have opted out tracking");
```
pchien: i think you mean the opposite in the message

HelloMixpanel/HelloMixpanelTests/MixpanelOptOutTests.m
```diff
@@ -134,7 +126,6 @@ - (void)testOptOutTrackingRegisterSuperPropertiesOnce {
 
 - (void)testOptOutWilSkipTimeEvent {
     [self.mixpanel optOutTracking];
```
pchien: oh i didn't notice last time. you're not actually timing anything here

HelloMixpanel/HelloMixpanelTests/MixpanelOptOutTests.m
```diff
@@ -191,8 +179,14 @@ - (void)testOptOutWillSkipFlushPeople
     
     NSMutableArray *peopleQueue = [NSMutableArray arrayWithArray:self.mixpanel.peopleQueue];
     [self.mixpanel optOutTracking];
-    [self waitForMixpanelQueues];
```
pchien: i think it's fine to wait here since it's more necessary. code will be cleaner too

HelloMixpanel/HelloMixpanelTests/MixpanelOptOutTests.m
```diff
@@ -212,8 +206,14 @@ - (void)testOptOutWillSkipFlushEvent
     
     NSMutableArray *eventsQueue = [NSMutableArray arrayWithArray:self.mixpanel.eventsQueue];
     [self.mixpanel optOutTracking];
-    [self waitForMixpanelQueues];
```
pchien: same here

HelloMixpanel/HelloMixpanelTests/MixpanelOptOutTests.m
```diff
@@ -0,0 +1,222 @@
...
+
+@interface MixpanelOptOutTests : MixpanelBaseTests
+
+@end
+
+@implementation MixpanelOptOutTests
+
+- (void)setUp
+{
+    [super setUp];
```
pchien: oh this looks unnecessary now

Mixpanel/Mixpanel.m
```diff
@@ -741,6 +813,36 @@ - (void)dispatchOnNetworkQueue:(void (^)(void))dispatchBlock {
...
+            self.people.unidentifiedQueue = [NSMutableArray array];
+            self.eventsQueue = [NSMutableArray array];
+            self.peopleQueue = [NSMutableArray array];
+            self.timedEvents = [NSMutableDictionary dictionary];
+        }
+        [self archive];
+    });
+}
+
+- (void)optInTracking
```
patedit: Missing a track call with `$opt_in` as the event name 👍

Mixpanel/Mixpanel.m
```diff
@@ -856,6 +962,11 @@ - (NSString *)eventBindingsFilePath
     return [self filePathFor:@"event_bindings"];
 }
 
+- (NSString *)optOutFilePath
+{
+    return [self filePathFor:@"optOut"];
```
patedit: so we don't create a new file for each param, we could use https://github.com/mixpanel/mixpanel-iphone/blob/master/Mixpanel/AutomaticEvents.m#L26 instead
ideally, some refactoring of the local storage would be amazing since it's all over the place right now

Mixpanel/Mixpanel.m
```diff
@@ -741,6 +813,36 @@ - (void)dispatchOnNetworkQueue:(void (^)(void))dispatchBlock {
     });
 }
 
+- (void)optOutTracking
```
patedit: We are missing a couple of things here: we need to call `deleteUser` and `clearCharges` in the `people` instance :)

Mixpanel/Mixpanel.m
```diff
@@ -765,7 +867,7 @@ - (void)startFlushTimer
 {
     [self stopFlushTimer];
     dispatch_async(dispatch_get_main_queue(), ^{
-        if (self.flushInterval > 0) {
+        if (self.flushInterval > 0 && ![self hasOptedOutTracking]) {
```
patedit: if the user opts in later during the session, this change will prevent events to be sent and that's no bueno

HelloMixpanel/HelloMixpanelTests/MixpanelOptOutTests.m
```diff
@@ -0,0 +1,213 @@
...
+
+- (void)testOptOutTrackingWillNotGenerateEventQueue
+{
+    stubTrack();
+    [self.mixpanel optOutTracking];
+    for (NSUInteger i = 0, n = 50; i < n; i++) {
+        [self.mixpanel track:[NSString stringWithFormat:@"event %lu", (unsigned long)i]];
+    }
+    [self waitForMixpanelQueues];
+    XCTAssertTrue([self.mixpanel.eventsQueue count] == 0, @"When opted out, people should not be queued");
```
patedit: you mean `events`?

HelloMixpanel/HelloMixpanelTests/MixpanelOptOutTests.m
```diff
@@ -0,0 +1,213 @@
...
+//  Created by Zihe Jia on 3/15/18.
+//  Copyright © 2018 Mixpanel. All rights reserved.
+//
+
+#import "MixpanelBaseTests.h"
+#import "MixpanelPrivate.h"
+#import "TestConstants.h"
+#import "MixpanelPeoplePrivate.h"
+
+@interface MixpanelOptOutTests : MixpanelBaseTests
```
patedit: We could add some other useful test cases. For example:
- Init the SDK with optOutTracking=false and make sure that no events are sent or in the queue (this test will fail right now - see comments below)
- make sure that when the user opts in, we have an `$opt_in` event in the queue only

Mixpanel/Mixpanel.m
```diff
@@ -741,6 +813,36 @@ - (void)dispatchOnNetworkQueue:(void (^)(void))dispatchBlock {
...
 }
 
+- (void)optOutTracking
+{
+    [self.optOutStatus setObject:@YES forKey:self.distinctId];
+    [self archiveOptOut];
+    dispatch_async(self.serialQueue, ^{
+        @synchronized (self) {
+            self.superProperties = [NSMutableDictionary dictionary];
+            self.people.distinctId = nil;
```
patedit: We should also remove `self.distinctId` !

Mixpanel/Mixpanel.m
```diff
@@ -374,6 +420,10 @@ - (void)identify:(NSString *)distinctId;
 
 - (void)identify:(NSString *)distinctId usePeople:(BOOL)usePeople;
 {
+    if ([self hasOptedOutTracking]) {
```
patedit: We should add this new distinct ID to the list of opt-out distinct ids as well

Mixpanel/Mixpanel.h
```diff
@@ -605,6 +633,27 @@ extern NSString *const MPNotificationTypeTakeover;
...
+ */
+- (void)optOutTracking;
+
+/*!
+ Opt in tracking.
+ 
+ This method is used to opt in tracking. This causes all events and people request to be sent back
+ to the Mixpanel server.
+ */
+- (void)optInTracking;
```
patedit: so we have parity with android what do you think if we create
`optInTrackingForDistinctId:(NSString *distinctId)` and `optInTrackingForDistinctId:(NSString *distinctId) withEventProperties:(NSDictionary *properties)` ?

Mixpanel/Mixpanel.m
```diff
@@ -151,6 +168,11 @@ - (instancetype)initWithToken:(NSString *)apiToken
         self.people = [[MixpanelPeople alloc] initWithMixpanel:self];
         [self setUpListeners];
         [self unarchive];
+        
+        if (optOutTracking) {
```
patedit: If a customer changes the default implementation from one version to the other, the end user will never be opt-in by default

HelloMixpanel/HelloMixpanelTests/MixpanelBaseTests.m
```diff
@@ -37,6 +37,7 @@ - (void)setUpMixpanel {
     self.mixpanel = [[Mixpanel alloc] initWithToken:kTestToken
                                       launchOptions:nil
                                    andFlushInterval:60];
+    [self deleteOptOutSettingsWithMixpanelInstance:self.mixpanel];
```
patedit: this shouldn't be needed if we delete it in the tear down

PR #770, opened 2018-03-23
Allow the user of Mixpanel to set a NSURLSession for custom configuration

YusufSoomro:
It is as the title says. Due to the nature of our company, we wish to protect any outgoing network requests with certificate pinning so we do not send out protected information.

We noticed you were using NSURLSessionConfiguration#defaultSessionConfiguration for the configuration, but there was no way to ensure that we would only be contacting an endpoint that we ourselves have authorized.

We'd like to know if you think we should be setting the NSURLSession a different way, or if you have better suggestions :)


YusufSoomro:
@pchien @zihejia Any thoughts on this?

PR #771, opened 2018-03-29
Concurrency improvements

timrozum:
### My issue

I ran into what I think may be a bug in the Mixpanel SDK. This is how I reproduce it:

1. Sign into app with an existing account.
2. The sign in changes the Mixpanel distinct ID.
3. Because I changed the distinct ID I notify our code that controls Branch SDK because per Branch documentation anytime Mixpanel distinct ID changes Branch SDK should be updated.
4. When the code that controls Branch SDK is notified it gets the current distinct ID from Mixpanel.
5. The distinct ID does not match the distinct ID of the new signed in user. Instead it is the previous distinct ID that Mixpanel SDK stored which is a problem.

### What I think the root cause is

Looking at the Mixpanel SDK code it looks like a concurrency issue. When I call `identify` the function does `dispatch_async(self.serialQueue, ^{` to add thread safety to stuff like `self.distinctId = distinctId;`. Because `dispatch_async` is used and not `dispatch_sync`, the `distinctId` will not get updated during the current run loop. This explains why my code that controls Branch SDK does not get the distinct ID for the new signed in user.

### A solution

I think this could be fixed by using `dispatch_sync` instead of `dispatch_async` when accessing data like `distinctId`. And then a `dispatch_sync` can be added to public `distinctId` accessor to truly make it thread safe. Right now I don't think read access to distinctId is completely safe because it doesn't use any dispatch. Is there a specific reason why `dispatch_async` is used? It seems like the dispatch is just for safety and there isn't long running work being done so `dispatch_sync` should be fine.

Also use of `@synchronized` seems redundant and I think could be removed and just rely on GCD.

Because replacing `dispatch_async` with `dispatch_sync` can change the order that code gets executed somebody that is familiar with this code would probably need to make sure there are no order of execution issues.

### DIFF
--- a/Mixpanel/Mixpanel.h
+++ b/Mixpanel/Mixpanel.h
@@ -522,7 +522,7 @@ extern NSString *const MPNotificationTypeTakeover;
 /*!
  Clears all stored properties and distinct IDs. Useful if your app's user logs out.
  */
-- (void)reset;
+- (void)resetWithCompletion:(void (^)(void))handler;
 
 /*!
  Uploads queued data to the Mixpanel server.
--- a/Mixpanel/Mixpanel.h
+++ b/Mixpanel/Mixpanel.h
@@ -65,6 +65,13 @@ extern NSString *const MPNotificationTypeTakeover;
 @property (atomic, readonly, strong) MixpanelPeople *people;
 
 /*!
+ The distinct ID of the current user that can be called from any queue.
+
+ See the documentation for distinctId below for more information.
+ */
+- (NSString *)safeDistinctId;
+
+/*!
  The distinct ID of the current user.
 
  A distinct ID is a string that uniquely identifies one of your users. By default, 
--- a/Mixpanel/Mixpanel.m
+++ b/Mixpanel/Mixpanel.m
@@ -140,6 +140,8 @@ static NSString *defaultProjectToken;
         self.serialQueue = dispatch_queue_create([label UTF8String], DISPATCH_QUEUE_SERIAL);
         NSString *networkLabel = [label stringByAppendingString:@".network"];
         self.networkQueue = dispatch_queue_create([networkLabel UTF8String], DISPATCH_QUEUE_SERIAL);
+        NSString *backgroundLabel = [label stringByAppendingString:@".background"];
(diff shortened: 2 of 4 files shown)


Mixpanel/Mixpanel.h
```diff
@@ -64,6 +64,13 @@ extern NSString *const MPNotificationTypeTakeover;
  */
 @property (atomic, readonly, strong) MixpanelPeople *people;
 
+/*!
+ The distinct ID of the current user that can be called from any queue.
+
+ See the documentation for distinctId below for more information.
+ */
+- (NSString *)safeDistinctId;
```
timrozum: Obviously there should only be one public distinct ID but I just added this one to minimize changes for initial feedback.

Mixpanel/Mixpanel.m
```diff
@@ -694,51 +706,48 @@ - (double)eventElapsedTime:(NSString *)event {
...
 
 - (void)dispatchOnNetworkQueue:(void (^)(void))dispatchBlock {
-    // so this looks stupid but to make [Mixpanel track]; [Mixpanel flush]; continue to have the track
-    // guaranteed to be part of the flush we need to make networkQueue stuff be dispatched on serialQueue
-    // first. still will allow serialQueue stuff to happen at the same time as networkQueue stuff just
-    // don't want to change track -> flush behavior that people may be relying on
-    dispatch_async(self.serialQueue, ^{
-        dispatch_async(self.networkQueue, dispatchBlock);
-    });
+    dispatch_async(self.networkQueue, dispatchBlock);
```
timrozum: I could be missing something but I think using `dispatch_async` allows this to be simplified.

timrozum:
I've realized that `flushWithCompletion` enables another solution that is pretty clean. After sign in for example I do some Mixpanel work then call `flushWithCompletion` and inside the closure I notify my Branch code to send a new distinct ID to Branch SDK. My Branch code gets the current distinct ID from Mixpanel which is now the correct value because of the flush. 

I'll use `flushWithCompletion` for now but I think it would be useful to look at the changes in this PR since I think they may simplify concurrency code and improve safety.

Issue #345, opened 2015-09-18
Carthage: dyld: Library not loaded: @rpath/Mixpanel.framework/Mixpanel


vandutran:
Got this error when running my app.

dyld: Library not loaded: @rpath/Mixpanel.framework/Mixpanel
Reason: image not found

Xcode 6.4, mixpanel 2.8.3, dragged carthage generated .framework file to project. No compile errors.

brennanMKE:
This happened to my project as well.

brennanMKE:
I resolved this problem with my project. I went to the General tab for the build target and added the Mixpanel framework in the Embedded Binaries section. I did find that it then doubled up the reference to Mixpanel in the frameworks so I removed it from both and added it again to Embedded Binaries and it only had one reference in both sections. Now the app runs without any errors.

@vandutran

samgreen:
@vandutran Is this still an issue for you?

vandutran:
I stopped using the Carthage framework after getting this error, now i'm importing the whole source code (.h, .m)

Haven't retried the carthage framework, but i'll give it another try later.

samgreen:
Bumping this to see if v2.9.0 resolves this @vandutran

vandutran:
Works now. Thanks.

vandutran:
I spoke too fast, got a crash at launch.

dyld: Library not loaded: @rpath/Mixpanel.framework/Mixpanel
Referenced from: /var/mobile/Containers/Bundle/Application/007B2C51-2BEA-466A-AFB9-A332C57192FB/xxx.app/xxx
Reason: image not found

Xcode 7

samgreen:
@vandutran Can you try an experiment and link to the framework weakly?

I'm not having that issue in our internal carthage test project.

vandutran:
I made the framework "Optional" in Linked Frameworks and Librairies and I get an exception when it reaches the Mixpanel.sharedInstance() line.

Adding the framework in "Embedded Binaries" like @brennanMKE suggested works. But I don't know if that's the right way to do things?

It's pretty simple to reproduce, you can just create a blank swift project, carthage mixpanel, and add it to the project.

jaredmixpanel:
I also ran into this issue and was able to fix it by adding a Copy Files Build Phase, setting its destination to Frameworks and adding the Mixpanel framework there (in addition to Link Binary with Libraries), like this: 
[...]

PR #778, opened 2018-04-24
Document required copying of framework

wbyoung:
Without doing this, the following error is produced when running an app
that links to the framework:

```
Dyld Error Message:
  Library not loaded: @rpath/Mixpanel.framework/Mixpanel
```

This is also noted in #345 with this same solution.


zihejia:
hi @wbyoung , this step should not be required if you follow the guide importing the project properly, but thanks anyway.

wbyoung:
Linking does not imply that Xcode should or will copy the framework. This _was_ required for me.  I am _very_ familiar with the build process in Xcode. It certainly does have a way of its own when dragging and dropping things to get it configured. So you can choose to do as you want, but I do know that it was required for me and I understand the reasons why as well. It wasn't just me guessing.

Issue #779, opened 2018-05-03
Mixpanel sharedInstanceWithToken:launchOptions:trackCrashes:automaticPushTracking: doesn't use declared boolean arguments


abotkin-cpi:
**Integration Method:** CocoaPods
**Xcode Version:** 9.2 (9C40b)
**Library Version:** 3.3.0
**Platform:** iOS
**Language:** Objective-C
**Description:** `-[Mixpanel sharedInstanceWithToken:launchOptions:trackCrashes:automaticPushTracking:]` does not use the declared boolean arguments for trackCrashes & automaticPushTracking but instead uses a hard-coded YES for both arguments (likely due to copy-paste).
**Expected Behavior:** Based off the documentation in Mixpanel.h, I would expect use of `-[Mixpanel sharedInstanceWithToken:launchOptions:trackCrashes:automaticPushTracking:]` with trackCrashes & automaticPushTracking set to NO woud result in both being turned off.

zihejia:
hi @abotkin-cpi, sorry about it, we have released a new version 3.3.1 with a fix to this problem.

PR #780, opened 2018-05-03
Ensure the trackCrashes & automaticPushTracking arguments are respected (#779)

abotkin-cpi:
This addresses the issue I raised in #779 where arguments that are declared are not respected in the method (likely due to a copy-paste error).  These changes should allow the method to match what the header documentation states.


zihejia:
Thanks @abotkin-cpi

PR #782, opened 2018-05-04
Harmonized usage of strong reference

philippec-ls:
Use a strong reference to self.mixPanel as done elsewhere in this method.


philippec-ls:
Something I noticed while integrating the latest Mixpanel for Xcode 9.3. compatibility. Guess we run with more warnings than the base project.

zihejia:
Thanks @philippec-ls

Issue #776, opened 2018-04-09
Watchdog crash


danielnagyx:
**Integration Method:** CocoaPods
**Xcode Version:** 9.2
**Library Version:** 3.2.6 
**Platform:** iOS 
**Language:** Objective-C
**Description:** Termination Reason: Namespace SPRINGBOARD, Code 0x8badf00d (main-thread watchdog crash [1])

We have observed several instances of our application crashing on older devices due to the main-thread watchdog killing the app. For example: in one instance  `[Mixpanel applicationDidEnterBackground:]` was blocking the main thread with `@synchronized(self)` because on a background thread (`com.mixpanel....`) the Mixpanel SDK was performing I/O operations also within a `@synchronized(self)` block. (`[Mixpanel track:properties:]` --> `[Mixpanel archiveEvents]` --> `[Mixpanel archiveObject:withFilePath:]`).

Running I/O operations within `@synchronized` blocks while synchronizing the main thread on the same object defeats the purpose of deferring I/O to a background thread.

**Expected Behavior:** 
The appropriate way of doing this would be to perform I/O outside of the synchronized block, so that it won't be possible to block the main thread. E.g.:

```
- (void)saveData {
    NSMutableData* dataToSave = [NSMutableData new];
    @synchronized(self) {
        ... // build the data
    }
    [dataToSave writeToFile:...];
}
```

[1] https://developer.apple.com/library/content/qa/qa1693/_index.html

tzahola:
Can confirm, this happened to me as well.

zihejia:
Thanks @danielnagyx and @tzahola, we will investigate on this and might create a PR for this.

zihejia:
hi @danielnagyx @tzahola , we have release v3.3.3 with a fix for this issue, thanks! Feel free to reopen it anytime if still have the problem.

PR #789, opened 2018-06-13
Fix the issue that applicationDidEnterBackground is blocking the main…

zihejia:
Fix the issue that applicationDidEnterBackground is blocking the main thread with @synchronized(self) inside `archive`, result in crash triggered by watchdog.

https://github.com/mixpanel/mixpanel-iphone/issues/776
@synchronized does not need to be inside `archive`, because in Mixpanel internally `archive` itself is always being called inside a serial queue.
Fix anywhere if `archive` not being called in a background thread.

### DIFF
--- a/Mixpanel/Mixpanel.m
+++ b/Mixpanel/Mixpanel.m
@@ -1514,9 +1514,9 @@ static void MixpanelReachabilityCallback(SCNetworkReachabilityRef target, SCNetw
     if (self.flushOnBackground) {
         [self flush];
     } else {
-        [self dispatchOnNetworkQueue:^{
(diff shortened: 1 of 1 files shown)


Mixpanel/Mixpanel.m
```diff
@@ -1526,15 +1514,19 @@ - (void)applicationWillResignActive:(NSNotification *)notification
     if (self.flushOnBackground) {
         [self flush];
     } else {
-        [self archive];
+        [self dispatchOnNetworkQueue:^{
```
patedit: why on the network queue and not the serial queue?

PR #792, opened 2018-07-26
hide keyboard while showing inapp-message

omerkolkanat:
When a takeover notification received while using the keyboard, keyboard stays on the window like the image below and doesn't disappear until to close the notification.

whatsapp image 2018-07-26 at 13 23 21


mesutdag:
+1

PR #793, opened 2018-09-13
Add new watch architecture

awe-mt:
The series 4 Apple Watch announced today uses a new architecture.

Also, the deployment target for watchOS must be increased to 3.0, otherwise uploads to the App Store will result in the following: 

> ERROR ITMS-90081: "This bundle is invalid. Applications built for more than one architecture require an iOS Deployment Target of 3.0 or later."


zihejia:
hi @awe-mt , thanks for providing this pr.

PR #798, opened 2018-09-21
Xcode 10 support

zihejia:


zihejia:
Will use a more consolidated PR https://github.com/mixpanel/mixpanel-iphone/pull/803

Issue #800, opened 2018-10-09
Can not track UITextField UIControlEventEditingDidEnd event on iOS12.


SecondDog:
The root cause is that in iOS12 UITextField method didMoveToWindow was **overwrite and it did not call its super method。** So when mixpanel try to hook UIControl Class's didMoveToWindow Method，the UITextField‘s didMoveToWindow was not hooked and called .That result in the event not actually binded and tracked.

Fix it like this:

in **UIControlBinding.m**
 initWithEventID:onPath:withControlEvent:andVerifyEvent:index:navTitle

add code:

```
......
if (controlEvent == UIControlEventEditingDidEnd ){
            [self setSwizzleClass:[UITextField class]];
}
......
```

zihejia:
Thanks so much @SecondDog, do you mind making a pull request for this fix?

SecondDog:
> Thanks so much @SecondDog, do you mind making a pull request for this fix?

OK 。 No Problem。

zihejia:
We have released v3.3.4 and has a fix for this. Thanks @SecondDog !

PR #801, opened 2018-10-10
Can not track UITextField UIControlEventEditingDidEnd event on iOS12. #800

SecondDog:
Change-Id: Ic23431e7afae65e15c4faf70d40ea6924af66db6

See this issue.

[mixpanel-iphone/issues#800](https://github.com/mixpanel/mixpanel-iphone/issues/800)

### DIFF
--- a/Mixpanel/MPUIControlBinding.m
+++ b/Mixpanel/MPUIControlBinding.m
@@ -77,6 +77,12 @@
         [self setSwizzleClass:[UIControl class]];
         _controlEvent = controlEvent;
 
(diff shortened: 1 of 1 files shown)


zihejia:
Hi @SecondDog , this seems not solving the problem completely, but it's inspiring enough. I will continue to investigate and provide a fix in the next version. Thanks again!

zihejia:
Hi @SecondDog , we have merged your change with some minor tweak to the master. Again, thanks so much for providing this.

Issue #41, opened 2012-12-11
Warning when compiling Mixpanel.m in Release build


ghost:
Mixpanel.m:337:12: Unused variable 'v'

'v' seems to be used only in an NSAssert, but in Xcode 4 assertions are blocked by default in Release builds. It would be nice if Mixpanel compiled cleanly in this case.

PR #805, opened 2018-10-19
sdk to set user and device id fields on events and people updates (#41)

dneralla:


Mixpanel/Mixpanel.m
```diff
@@ -1145,7 +1157,14 @@ - (void)unarchiveProperties
 {
     NSDictionary *properties = (NSDictionary *)[Mixpanel unarchiveFromFile:[self propertiesFilePath] asClass:[NSDictionary class]];
     if (properties) {
-        self.distinctId = properties[@"distinctId"] ?: [self defaultDistinctId];
+        self.distinctId = properties[@"distinctId"];
+        self.userId     = properties[@"userId"];
+        self.anonymousId = properties[@"anonymousId"];
+        if(!self.distinctId) {
```
zihejia: Nit: if (!self.distinctId) {

dneralla: done

PR #810, opened 2018-10-26
SDK changes to support group- and account-level analytics

yuanweiz:
DG-188
SDK changes to support group- and account-level analytics


zihejia:
use https://github.com/mixpanel/mixpanel-iphone/pull/817 instead

PR #817, opened 2018-11-27
SDK changes to support group and account level analytics

zihejia:
SDK changes to support group and account level analytics


yas375:
Is there an article or documentation about groups in Mixpanel? I'm curious what it is.

PR #826, opened 2019-01-30
Add support for display triggers in InApp notifications

madster26:
**Event Triggered InApps**

Currently users have no control over when an InApp notification shows up. With this release users can now control when an InApp gets displayed based on an event that they track within their app. This "trigger" event is defined during message creation. You can additionally filter the event based on properties that are tracked along with the event for even finer controls.

### DIFF
--- a/Mixpanel/MPDisplayTrigger.m
+++ b/Mixpanel/MPDisplayTrigger.m
@@ -1,6 +1,7 @@
 #import "MPLogger.h"
 #import "MPDisplayTrigger.h"
 #import "Mixpanel.h"
+#import "Mixpanel/Mixpanel-Swift.h"
 
 NSString * const ANY_EVENT = @"$any_event";
 
@@ -18,7 +19,7 @@ NSString * const ANY_EVENT = @"$any_event";
             event = @"";
         }
         
-        _jsonObject = object;
+        _rawJSON = object;
         _event = event;
         _selector = object[@"selector"];
     }
@@ -32,10 +33,12 @@ NSString * const ANY_EVENT = @"$any_event";
     }
     
     NSString *eventName = event[@"event"];
-    if ([eventName caseInsensitiveCompare:@""] == NSOrderedSame ||
-        [eventName caseInsensitiveCompare:ANY_EVENT] == NSOrderedSame ||
-        [eventName caseInsensitiveCompare:[self event]] == NSOrderedSame) {
-            return YES;
+    if ([eventName compare:ANY_EVENT] == NSOrderedSame ||
+        [eventName compare:[self event]] == NSOrderedSame) {
+        if ([_selector count] > 0) {
+            return [SelectorEvaluator evaluateWithSelector:_selector properties:event[@"properties"]];
+        }
+        return YES;
     }
     
     return NO;
--- a/Mixpanel/MPDisplayTrigger.m
+++ b/Mixpanel/MPDisplayTrigger.m
@@ -1,0 +1,44 @@
+#import "MPLogger.h"
+#import "MPDisplayTrigger.h"
+#import "Mixpanel.h"
+
+NSString * const ANY_EVENT = @"$any_event";
+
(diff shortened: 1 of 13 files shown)


Mixpanel/MPDisplayTrigger.m
```diff
@@ -0,0 +1,45 @@
...
+        
+        NSString *event = object[@"event"];
+        if ([event isEqual:[NSNull null]]) {
+            event = @"";
+        }
+        
+        _jsonObject = object;
+        _event = event;
+        _selector = object[@"selector"];
+
```
zihejia: NIT: extra line here

Mixpanel/MPDisplayTrigger.m
```diff
@@ -0,0 +1,45 @@
...
+        _selector = object[@"selector"];
+        
+    }
+    
+    return self;
+}
+
+- (BOOL)matchesEvent:(NSDictionary *)event {
+    if (event == nil) {
+        return false;
```
zihejia: ObjC Coding style: false -> NO, true -> YES

Mixpanel/MPDisplayTrigger.m
```diff
@@ -0,0 +1,45 @@
...
+    }
+    
+    NSString *eventName = event[@"event"];
+    if ([eventName caseInsensitiveCompare:@""] == NSOrderedSame ||
+        [eventName caseInsensitiveCompare:ANY_EVENT] == NSOrderedSame ||
+        [eventName caseInsensitiveCompare:[self event]] == NSOrderedSame) {
+            return true;
```
zihejia: `true` -> `YES` the same as above and across the PR

Mixpanel/MPNotification.m
```diff
@@ -86,11 +95,27 @@ - (instancetype)initWithJSONObject:(NSDictionary *)object {
...
         _imageURL = imageURL;
         _image = nil;
+        _displayTriggers = parsedDisplayTriggers;
     }
 
     return self;
 }
 
+- (BOOL)hasDisplayTriggers {
+    return _displayTriggers != nil && [_displayTriggers count] > 0;
```
zihejia: ObjC Coding style: prefer `self.displayTriggers` than `_displayTriggers` except in init. the same comment to all occurrences in this PR.

Mixpanel/Mixpanel.m
```diff
@@ -28,7 +29,7 @@
 #error The Mixpanel library must be compiled with ARC enabled
 #endif
 
-#define VERSION @"3.3.9"
+#define VERSION @"3.4.0"
```
zihejia: version bump is not needed in the PR, this will be added after the PR being merged.

Mixpanel/MixpanelPrivate.h
```diff
@@ -112,6 +112,7 @@
 @property (nonatomic) BOOL hasAddedObserver;
 @property (nonatomic, strong) NSNumber *automaticEventsEnabled;
 @property (nonatomic, strong) NSArray *notifications;
+@property (nonatomic, strong) NSArray *triggeredNotifications;
```
zihejia: using (nonatomic, copy) is recommended, since you are here, could you fix it for notifications to @property (nonatomic, copy) as well? In objc, use copy for classes that are part of a class cluster that have mutable/immutable pairs; such as NSString/NSMutableString NSArray/NSMutableArray NSDictionary/NSMutableDictionary NSSet/NSMutableSet

Mixpanel/SelectorEvaluator.swift
```diff
@@ -0,0 +1,566 @@
...
+            default:
+                return false
+            }
+        }
+        return false
+    }
+    
+    class func evaluateNumber(node: [String: Any], properties: Properties?) -> Double? {
+        guard let op = node[OPERATOR_KEY] as? String, op == NUMBER_OPERATOR else {
+            MPLog
```
zihejia: typo here

PR #833, opened 2019-03-20
Add native implementation for SelectorEvaluator

madster26:


Mixpanel/MPDisplayTrigger.m
```diff
@@ -1,7 +1,7 @@
 #import "MPLogger.h"
 #import "MPDisplayTrigger.h"
+#import "SelectorEvaluator.h"
 #import "Mixpanel.h"
-#import "Mixpanel/Mixpanel-Swift.h"
 
 NSString * const ANY_EVENT = @"$any_event";
```
zihejia:
```suggestion
static NSString *const ANY_EVENT = @"$any_event";
```
static is preferred, since it means the const is specific and internal to this .m file only

Mixpanel/SelectorEvaluator.h
```diff
@@ -0,0 +1,101 @@
+//
+//  SelectorEvaluator.h
+//  Mixpanel
+//
+//  Created by Madhu Palani on 3/15/19.
+//  Copyright © 2019 Mixpanel. All rights reserved.
+//
+#import <Foundation/Foundation.h>
+@interface MPBoolean : NSObject
```
zihejia: nit: add an additional line here

Mixpanel/SelectorEvaluator.h
```diff
@@ -0,0 +1,101 @@
...
+
+@property (nonatomic, readonly) BOOL value;
+
+- (instancetype)init:(BOOL)value;
+
+@end
+
+@interface SelectorEvaluator : NSObject
++(NSDate*) getCurrentDate;
```
zihejia: nit: add an additional line here

Mixpanel/SelectorEvaluator.h
```diff
@@ -0,0 +1,101 @@
...
++(NSDateFormatter*) getDateFormatter;
```
zihejia: nit: + (NSDateFormatter *)getDateFormatter;
also in objective c, for naming methods, normally we don't use `get` here.
https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingMethods.html

Mixpanel/SelectorEvaluator.m
```diff
@@ -0,0 +1,761 @@
...
+//  SelectorEvaluator.m
+//  Mixpanel
+//
+//  Created by Madhu Palani on 3/15/19.
+//  Copyright © 2019 Mixpanel. All rights reserved.
+//
+
+#import "SelectorEvaluator.h"
+
+NSString* const errDomain = @"SelectorEvaluatorError";
```
zihejia: nit: same as above, static NSString *const ...

Mixpanel/SelectorEvaluator.m
```diff
@@ -0,0 +1,761 @@
...
+NSString* const LESS_THAN_OPERATOR = @"<";
+NSString* const LESS_THAN_EQUAL_OPERATOR = @"<=";
+// Unary operators
+NSString* const NOT_OPERATOR = @"not";
+NSString* const DEFINED_OPERATOR = @"defined";
+NSString* const NOT_DEFINED_OPERATOR = @"not defined";
+// Special words
+NSString* const NOW_LITERAL = @"now";
+
+int const LEFT = 0;
```
zihejia: in Objective C, we usually want to use `NSInteger` rather than `int`, more detail: https://stackoverflow.com/questions/4445173/when-to-use-nsinteger-vs-int

zihejia:
```suggestion
static const NSInteger kLEFT = 0;
```

Mixpanel/SelectorEvaluator.m
```diff
@@ -0,0 +1,761 @@
...
+    
+    return [self isEqualToMPBoolean:object];
+}
+
+@end
+
+@implementation SelectorEvaluator
+
++ (NSDate *) getCurrentDate {
+    return [NSDate date];
```
zihejia: you can use [NSDate date] directly, in ObjC, it's quite standard.

Mixpanel/SelectorEvaluator.m
```diff
@@ -0,0 +1,761 @@
...
+        [formatter setDateFormat:@"yyyy-MM-dd'T'HH:mm:ss"];
+        [formatter setLocale:[[NSLocale alloc] initWithLocaleIdentifier:@"en_US_POSIX"]];
+    });
+    return formatter;
+}
+
++ (NSError *) getError:(NSString *) errDesc {
+    NSDictionary *info = @{
+       NSLocalizedDescriptionKey: NSLocalizedString(errDesc, nil),
+       NSLocalizedFailureReasonErrorKey: NSLocalizedString(errDesc, nil),
```
zihejia: if they are the same, you can just keep NSLocalizedDescriptionKey.

Mixpanel/SelectorEvaluator.m
```diff
@@ -0,0 +1,761 @@
...
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        formatter = [[NSDateFormatter alloc] init];
+        [formatter setDateFormat:@"yyyy-MM-dd'T'HH:mm:ss"];
+        [formatter setLocale:[[NSLocale alloc] initWithLocaleIdentifier:@"en_US_POSIX"]];
+    });
+    return formatter;
+}
+
++ (NSError *) getError:(NSString *) errDesc {
```
zihejia: better rename it without `get`

Mixpanel/SelectorEvaluator.m
```diff
@@ -0,0 +1,761 @@
...
+    NSDictionary *info = @{
+       NSLocalizedDescriptionKey: NSLocalizedString(errDesc, nil),
+       NSLocalizedFailureReasonErrorKey: NSLocalizedString(errDesc, nil),
+       NSLocalizedRecoverySuggestionErrorKey: NSLocalizedString(errDesc, nil),
+   };
+    
+    return [[NSError alloc] initWithDomain:errDomain code:-1 userInfo:info];
```
zihejia: better replace -1 with a const

Mixpanel/SelectorEvaluator.m
```diff
@@ -0,0 +1,761 @@
...
+}
+
++ (NSNumber *) evaluateNumber:(NSDictionary *)node properties:(NSDictionary *)properties withError:(NSError *__autoreleasing *)error {
+    if (![node objectForKey:OPERATOR_KEY] || ![node[OPERATOR_KEY] isEqualToString:NUMBER_OPERATOR]) {
+        if (error) {
+            *error = [self getError:@"invalid operator: number"];
+        }
+        return nil;
+    }
+    if (![node objectForKey:CHILDREN_KEY]  || !([(NSArray *)node[CHILDREN_KEY] isKindOfClass:[NSArray class]] && ([(NSArray *)node[CHILDREN_KEY] count] == 1))) {
```
zihejia: nit: line is getting too long

Mixpanel/SelectorEvaluator.m
```diff
@@ -0,0 +1,761 @@
...
+    id l = [self evaluateNode:node[CHILDREN_KEY][LEFT] properties:properties withError:error];
+    if (error && *error) {
+        return nil;
+    }
+    id r = [self evaluateNode:node[CHILDREN_KEY][RIGHT] properties:properties withError:error];
+    if (error && *error) {
+        return nil;
+    }
+    BOOL b = NO;
+    if ([(NSString *) l isKindOfClass:[NSString class]] && [(NSString *) r isKindOfClass:[NSString class]]) {
```
zihejia: nit: (NSString *)l (without space)

Mixpanel/SelectorEvaluator.m
```diff
@@ -0,0 +1,761 @@
...
+        }
+        if ([op isEqualToString:MOD_OPERATOR]) {
+            if (rd == 0) {
+                return nil;
+            }
+            if (ld == 0) {
+                return [NSNumber numberWithDouble:0];
+            }
+            if ((ld < 0 && rd > 0) || (ld > 0 && rd < 0)) {
+                return [NSNumber numberWithDouble:-(floor(ld/rd) * rd - ld)];
```
zihejia:
```suggestion
                return [NSNumber numberWithDouble:-(floor(ld / rd) * rd - ld)];
```

Mixpanel/SelectorEvaluator.m
```diff
@@ -0,0 +1,761 @@
...
+        return l >= r;
+    }
+    if ([op isEqualToString:LESS_THAN_OPERATOR]) {
+        return l < r;
+    }
+    if ([op isEqualToString:LESS_THAN_EQUAL_OPERATOR]) {
+        return l <= r;
+    }
+    
+    return false;
```
zihejia:
```suggestion
    return NO;
```

Mixpanel/SelectorEvaluator.m
```diff
@@ -0,0 +1,761 @@
...
+    id l = [self evaluateNode:node[CHILDREN_KEY][LEFT] properties:properties withError:error];
+    if (error && *error) {
+        return nil;
+    }
+    id r = [self evaluateNode:node[CHILDREN_KEY][RIGHT] properties:properties withError:error];
+    if (error && *error) {
+        return nil;
+    }
+    
+    BOOL b = false;
```
zihejia:
```suggestion
    BOOL b = NO;
```

Mixpanel/SelectorEvaluator.m
madster26: The reason this function exists is so that I can override the default date in tests

madster26: I have added some comments

PR #871, opened 2019-11-14
Update Mixpanel.podspec

### LINKED ISSUE #870: iOS Deployment target issue
https://github.com/mixpanel/mixpanel-iphone/blob/5be3385af99f3c4cd4c1971b834e45c41576f48e/Mixpanel.podspec#L12

Some features in the Mixpanel code are using `UNUserNotificationCenter` which is only available on iOS 10.0 and above, so deployment target needs to be adjusted to 10.0 and above, otherwise there are warnings in the code.

Screen Shot 2019-11-14 at 3 58 59 PM

gligorkot:
Fixes #870

### DIFF
--- a/Mixpanel.podspec
+++ b/Mixpanel.podspec
@@ -9,7 +9,7 @@ Pod::Spec.new do |s|
   s.requires_arc = true
   s.libraries = 'icucore'
   s.swift_version = '4.2'
-  s.ios.deployment_target = '8.0'
+  s.ios.deployment_target = '10.0'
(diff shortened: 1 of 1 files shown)


zihejia:
hi @gligorkot , the deployment target needs to be changed as well. I will take care of it, thanks for bringing this up.

gligorkot:
@zihejia good point, sorry I tried a quick fix, but not enough.

gligorkot:
#892 is my second attempt at getting this fixed.

PR #879, opened 2020-02-04
Improve push tracking

jbwyme:
- Adds `trackPushNotificationEventFromRequest` to track events using the `token` and `distinct_id` we receive from the push payload
- Adds a `$push_notification_received` event that is tracked when the device receives the notification
- Some other minor cleanup


HelloMixpanel/HelloMixpanel/AppDelegate.m
```diff
@@ -3,7 +3,7 @@
 #import "ViewController.h"
 
 // IMPORTANT!!! replace with your api token from https://mixpanel.com/account/
-#define MIXPANEL_TOKEN @"YOUR_MIXPANEL_PROJECT_TOKEN"
+#define MIXPANEL_TOKEN @"6888bfdec29d84ab2d36ae18c57b8535"
```
aecast: Do we still want an actual token here instead of a placeholder?

jbwyme: remove

PR #888, opened 2020-03-27
update docs for ID merge

zihejia:


Mixpanel/Mixpanel.h
```diff
@@ -412,24 +412,24 @@ extern NSString *const MPNotificationTypeTakeover;
...
+ Mixpanel will use the IFV as the default local distinct ID.
 
  If we are unable to get an IFA or IFV, we will fall back to generating a
  random persistent UUID. If you want to always use a random persistent UUID
  you can define the <code>MIXPANEL_RANDOM_DISTINCT_ID</code> preprocessor flag
  in your build settings.
 
  For tracking events, you do not need to call <code>identify:</code> if you
- want to use the default.  However, <b>Mixpanel People always requires an
+ want to use the default.  However, <b>Mixpanel User profiles always require an
```
cassiegamm: remove " if you want to use the default."

cassiegamm: (as context: with id merge enabled, you can't control the resolved distinct_id -- you used to be able to which is why the sentence was worded this way)

PR #875, opened 2019-12-19
Fix compile under catalyst (iosmac)

tmm1:

PR #896, opened 2020-04-28
Add support for catalyst(iosmac) 

zihejia:
credit to https://github.com/mixpanel/mixpanel-iphone/pull/875, thanks so much @tmm1


tmm1:
:+1:

PR #902, opened 2020-05-23
os_log* wrappers need to be always inlined

Kentzo:
The os_log* functions will not capture context-specific information unless inlined.


zihejia:
Thanks @Kentzo

PR #907, opened 2020-07-08
Remove using IFA as distinct id

zihejia:
- After iOS 14, accessing IFA will require an app developer to obtain an end-user's consent via an opt-in. An app will need to obtain permission to access this unique identifier, it imposes some challenges for us like caching, permissions and usage of new frameworks and doesn't add a lot of value. We will default to random UUID instead. If you want to  use a unique persistent UUID(IFV), you can define the <code>MIXPANEL_UNIQUE_DISTINCT_ID</code>  preprocessor flag in your build settings.
- Customers can still set/identify the distinct_id as IFA or $ios_ifa property by themselves if they have access to the advertising identifier.


Mixpanel/Mixpanel.h
```diff
@@ -407,26 +407,14 @@ extern NSString *const MPNotificationTypeTakeover;
...
- app, Mixpanel will use the IFA as the default local distinct ID. If you have
- AdSupport installed but still don't want to use the IFA, you can define the
- <code>MIXPANEL_NO_IFA</code> preprocessor flag in your build settings, and
- Mixpanel will use the IFV as the default local distinct ID.
-
- If we are unable to get an IFA or IFV, we will fall back to generating a
- random persistent UUID. If you want to always use a random persistent UUID
- you can define the <code>MIXPANEL_RANDOM_DISTINCT_ID</code> preprocessor flag
- in your build settings.
+ Mixpanel use a random persistent UUID as the default local distinct ID.
```
patedit, line 410:
```suggestion
 Mixpanel uses a randomly generated persistent UUID as the default local distinct ID.
```

Issue #909, opened 2020-08-09
Crashes in encodeArrayForAPI: and convertFoundationTypesToJSON: under iOS 13


PavelGnatyuk:
**Integration Method:** CocoaPods
**Xcode Version:** Xcode 11.6
**Library Version:** 
**Platform:** iOS 
**Language:** Objective-C
**Description:** 
**Expected Behavior:** 

Crash happens in the encodeArrayAsJSONData call or inside it in encodeArrayAsJSONData (exactly in the conversion to JSON):

```
+ (NSString *)encodeArrayForAPI:(NSArray *)array {
    NSData *data = [MPNetwork encodeArrayAsJSONData:array];  // Crash in this line. EXC_BAD_ACCESS KERN_INVALID_ADDRESS
    return [MPNetwork encodeJSONDataAsBase64:data];
} 

```

```
+ (NSData *)encodeArrayAsJSONData:(NSArray *)array {
    NSError *error = NULL;
    NSData *data = nil;
    @try {
        data = [NSJSONSerialization dataWithJSONObject:[self convertFoundationTypesToJSON:array]
                                               options:(NSJSONWritingOptions)0
                                                 error:&error];  //crash in this line EXC_BAD_ACCESS KERN_INVALID_ADDRESS
    }
    @catch (NSException *exception) {
        MPLogError(@"exception encoding api data: %@", exception);
    }
    
    if (error) {
        MPLogError(@"error encoding api data: %@", error);
    }
    
    return data;
}
```

zihejia:
hi @PavelGnatyuk , I'm very sorry for the late response. Could you provide more information? This will help us a lot for narrowing it down.
1. Do you have a full crash log showing all threads?
2. Did it happen only in iOS 13, what is the percentage of the crash if you have the data?
3. Do you use crashlytics, do you see any other error logs? 
4. Any chance you can reproduce it?
5. which SDK version you are using.
6. Any other information you'd like to share would be super helpful.

In the meantime, I'm thinking of two possible causes for this crash but need more information to deep dive further. 
1) It might related to the iOS 13 bug for `encodeArrayAsJSONData:` if the input `array` in `encodeArrayAsJSONData:(NSArray *)array` contains special characters, see radar https://openradar.appspot.com/radar?id=4964054569320448  (NSJSONSerialization crashes on iOS 13)
2) There is a race condition in the input `array` accessed by other threads trying to mutating it.

zihejia:
[...]

PR #910, opened 2020-09-24
Fix NSJSONSerialization crash in iOS 13

zihejia:
This is to address https://github.com/mixpanel/mixpanel-iphone/issues/909 (Crashes in encodeArrayForAPI: and convertFoundationTypesToJSON: under iOS 13)

In iOS 13 there is a bug for `encodeArrayAsJSONData:` if the input `array` in `encodeArrayAsJSONData:(NSArray *)array` contains corrupted characters(not valid UTF-8), it will lead to a crash. see radar https://openradar.appspot.com/radar?id=4964054569320448  (NSJSONSerialization crashes on iOS 13) The fix is to check whether the string is a valid UTF-8 before sending to Mixpanel. Mixpanel backend can only handle UTF-8 string.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ before_install:
   - gem install fourflusher
   - gem uninstall cocoapods --force -a -q -I -x
   - gem install cocoapods -v 1.6.1
-osx_image: xcode11
+osx_image: xcode12
 branches:
   only:
   - master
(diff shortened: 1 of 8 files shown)


freak4pc:
Hey @zihejia,
We saw a couple of crashes around archiving, is this related to this by any chance? 

image

freak4pc:
It's on 3.6.2.

freak4pc:
Opened #912. Thank you!

Issue #915, opened 2020-11-26
Crash & Data Race in [MixPanel archiveEvents]


AndreasVerhoeven:
**Integration Method:** CocoaPods
**Xcode Version:** 12.2
**Library Version:**  3.6.3 (also in main)
**Platform:** iOS
**Language:** Objective-C
**Description:**  Crash in `[MixPanel archiveEvents]`
**Expected Behavior:**  No crashes

Our users are getting crashes in `[MixPanel archiveEvents]`. A quick look in the code seems to reveal a simple data race:

```
- (void)archiveEvents
{
    NSString *filePath = [self eventsFilePath];
    MPLogInfo(@"%@ archiving events data to %@: %@", self, filePath, self.eventsQueue);
    dispatch_sync(self.archiveQueue, ^{
        NSArray *shadowEventsQueue =  [self.eventsQueue copy];
        if (![self archiveObject:shadowEventsQueue withFilePath:filePath]) {
            MPLogError(@"%@ unable to archive event data", self);
        }
    });
}
```

`self.eventsQueue` is copied here on the `archiveQueue`, while it's also accessed simultaneously on other queues. Without any locking here. [NSArray copy] is not an atomic operation, this needs to be protected by a lock or moved out of the queue.

**This is a simple data race**. Internally, `[NSArray copy]` does something like:
```
NSRange range = NSRangeMake(0, self.count);
[[NSArray alloc] initWithArray:self range:range copyItems:NO];
```
So, it's easy to see how this results in a data race and a crash: If the queue is flushed on the `dispatchOnNetworkQueue` in between the self.count and the copy of items, the copy thinks there are more items and thus crashes. 

This seems to happen with a lot of other data queue accesses as well, by the way. There seem to be several serial queues accessing the same data, which kind of defeats the point of using serial queues to protect data?

Crashing thread:
```
Thread 10 Crashed:
0   CoreFoundation                  0x33200c878         <redacted>
1   libobjc.A.dylib                 0x37b41fc50         objc_exception_throw
2   CoreFoundation                  0x33207ce1c         <redacted>
3   CoreFoundation                  0x331eeb3b8         <redacted>
4   SegmentAnalytics                0x103633efc         __25-[Mixpanel archiveEvents]_block_invoke
5   libdispatch.dylib               0x19d25cdb0         <redacted>
6   libdispatch.dylib               0x19d26b428         <redacted>
[...]

PR #917, opened 2020-12-17
Fix crash when archive caused by race condition

zihejia:
This is to address the issue: https://github.com/mixpanel/mixpanel-iphone/issues/915
[NSMutable copy] is not thread-safe, need better write access protection.


jaredmixpanel (approved):
lgtm

PR #918, opened 2020-12-17
temporarily remove  monitoring the radio change due to a bug from Apple

zihejia:
Temporarily remove the ability to monitor the radio change due to a crash issue that might relate to the API from Apple
https://openradar.appspot.com/46873673


jaredmixpanel (commented):
👍

PR #923, opened 2021-01-23
update deprecated apis

jaredmixpanel:


Mixpanel/Mixpanel.m
```diff
@@ -1351,8 +1351,21 @@ - (void)archiveOptOut
 - (BOOL)archiveObject:(id)object withFilePath:(NSString *)filePath
 {
     @try {
-        if (![NSKeyedArchiver archiveRootObject:object toFile:filePath]) {
-            return NO;
+        if (@available(iOS 11, macOS 10.13, tvOS 11, watchOS 4, *)) {
+            NSError *error = nil;
+            NSData *data = [NSKeyedArchiver archivedDataWithRootObject:object requiringSecureCoding:NO error:&error];
+            if (error) {
+                MPLogError(@"%@ got error while archiving data: %@", self, error);
```
zihejia, line 1358: return `NO`?

jaredmixpanel: hmmm, we `return NO` below `if (!data)`... so if `data` is non-nil and there is an error do we want to `return` before we write that data to file?

zihejia: yeah, if there is an error, the data will most likely be nil or corrupted. you can also do `if (error || !data)`

jaredmixpanel: changed to `if (error || !data)`

zihejia (approved):

PR #942, opened 2021-06-15
Set `automaticPushTracking` and `trackCrashes` to NO by default when init the SDK

zihejia:
We are deprecating Messaging and Mobile A/B testing features([more details](https://mixpanel.com/blog/why-were-sunsetting-messaging-and-experiments/)). Starting from the next version, we will not automatically track pushes sent from Mixpanel and crashes by default. However, you can still opt-in them by calling 
```
+ (Mixpanel *)sharedInstanceWithToken:(NSString *)apiToken launchOptions:(nullable NSDictionary *)launchOptions trackCrashes:(BOOL)trackCrashes automaticPushTracking:(BOOL)automaticPushTracking;
```
and set `trackCrashes` and `automaticPushTracking` to YES.


jaredmixpanel (approved):
lgtm

PR #948, opened 2021-10-03
NSSecureCoding Implementation

davidhodge:
**What:**

This patch puts Mixpanel in conformance with NSSecureCoding and should stop warnings (and possibly some crashes) about non-compliance.

**Why:**

Warnings were getting concerning and possible security matters are important to address. Also, there appears to possibly be a crasher with iOS 15.1 when not using secure coding. This last part is still speculative and I am investigating.

**How:**

This implementation, I believe, conforms to the "letter of the law" of NSSecureCoding, and makes progress on the "spirit" of the law. The change adds conformance to NSSecureCoding by adding "+ (BOOL)supportsSecureCoding", use of "-decodeObjectOfClass:forKey:", and changing the protocol from NSCoding to NSSecureCoding.

I'm implementing this to reduce warnings about lack of NSSecureCoding and to investigate a possible fix to a NSCoding crash I'm seeing on iOS 15.1 beta.

**Possible issues**
1. I applied this to master and only later realized I was unsure about the branching strategy and the use of 4.0.0.beta. It's possible I messed up in such a way this needs to be redone. Presumably anyone reading this at mixpanel understands the branching situation better than I.
2. Someone who understands the implementation better should probably clamp down more on MPTweakValue given that it seems that may just be intended to store a basic type.
3. This change is, as of this moment, untested other than compilation. I'm going to push out a TestFlight build of my app that uses it to try it out shortly.

### DIFF
--- a/Mixpanel/MPEventBinding.h
+++ b/Mixpanel/MPEventBinding.h
@@ -9,7 +9,7 @@
 #import <Foundation/Foundation.h>
 #import "MPObjectSelector.h"
 
-@interface MPEventBinding : NSObject <NSCoding>
+@interface MPEventBinding : NSObject <NSSecureCoding>
 
 @property (nonatomic) NSUInteger ID;
 @property (nonatomic, copy) NSString *name;
--- a/Mixpanel/MPEventBinding.m
+++ b/Mixpanel/MPEventBinding.m
@@ -87,14 +87,18 @@
 
 #pragma mark -- NSCoder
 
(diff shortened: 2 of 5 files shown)


davidhodge:
okay I am now a bit thrown as I don't see these classes at all in 4.0.0.beta. Perhaps this PR is un-needed?

zihejia:
yes @davidhodge , the pr is not needed but thanks so much for your contribution!

PR #954, opened 2021-11-11
Remove alias check when calling identify

argenisf:
In parity with the swift library, removing the check on existing alias when identify is called since the integration with Braze can prevent identify from working correctly. Also not relevant with ID merge.


zihejia (approved):
Thanks @argenisf

Issue #949, opened 2021-10-07
$carrier property can be reported as `nil` when there is multiple SIMs support


4hakke:
**Integration Method:** CocoaPods
**Xcode Version:** 12.5.1
**Library Version:** 
**Platform:** iOS
**Language:** Swift
**Testen on:** iPhone SE (2nd gen) iOS 13.5.1 and iPhone 12 Pro iOS 15.0
**Description:** 
We noticed that number of events where `$carrier` value is `underfined` was increased. After some investigation of `collectAutomaticProperties` method it was found that carrier name and all other properties retrieved from `telephonyInfo` are `nil` for devices. Further investigation showed that correct information are still provide by `telephonInfo` object but wrong object (always the first one aka `[[telephonyInfo serviceSubscriberCellularProviders] allValues].firstObject;`) were taken.

Down below you can see a console output for `[[telephonyInfo serviceSubscriberCellularProviders] allValues]` that proves needed information is there:

![Screenshot 2021-10-07 at 10 58 48](https://user-images.githubusercontent.com/15330332/136358119-88cfaf2d-6729-4d27-8776-f38800e32110.png)

I suppose that reason for that is devices that supports multiple sims (nano-Sim and esim) and only one of them is setup. In that case sim that are not setup still can be retrieved but are nil. So my suggestion to cover this case is to enhance logic for picking proper carrier (`CTCarrier`) object e.g. picking first where carrier name is not nil if any. 

**Expected Behavior:** 
If device has at least one carrier properly setup it should take object that has `name` information.

Happy to answer any additional questions or provide more insights. Thanks.

zihejia:
hi @4hakke , sorry for the late response and thanks so much for bringing this up. Do you mind opening a PR for this?

4hakke:
Hey @zihejia, sure thing, will make change and open PR soon.

zihejia:
hi @4hakke , It's been included in [v3.9.1](https://github.com/mixpanel/mixpanel-iphone/releases/tag/v3.9.1 ). Thank you so much!

PR #955, opened 2021-11-12
Carrier name retrieval

4hakke:
**Overview**
This PR updates logic for retrieving **$carrier** property to try to find the carrier that has a name that is not empty. 
More details on the reasoning behind the following change are described in this issue https://github.com/mixpanel/mixpanel-iphone/issues/949

**Changes**
In order to reduce the number of events where **$carrier** is `undefined` now it iterates over a collection of carriers and tries to find some that name is not empty. Once the proper carrier is found it breaks the loop. In case all carriers don't have a name it uses the first object as a fallback to be in line with previous logic.


zihejia (approved):
Thank you so much! @4hakke

Issue #970, opened 2022-01-04
Where do we specify preprocessor MIXPANEL_UNIQUE_DISTINCT_ID when we add mixpanel using SPM


tc-sgupta:
**Integration Method:** Swift Package Manager
**Xcode Version:** 13.2.1
**Library Version:** 4.0.0
**Platform:** iOS
**Language:** Objective-C
**Description:** Where do we specify preprocessor MIXPANEL_UNIQUE_DISTINCT_ID when we add mixpanel using SPM
**Expected Behavior:**

tc-sgupta:
We specify using cocoapods like this
```
post_install do |installer|
    installer.pods_project.targets.each do |target|
        target.build_configurations.each do |config|
            if ['Mixpanel'].include? target.name
                config.build_settings['GCC_PREPROCESSOR_DEFINITIONS'] ||= ['$(inherited)', 'MIXPANEL_UNIQUE_DISTINCT_ID=1']
            end
        end
    end
end
```

tc-sgupta:
@zihejia Could you please help us, how we can execute this code https://github.com/mixpanel/mixpanel-iphone/blob/b213461deae5356be7e917f1addbb3ebffe26b37/Sources/Mixpanel.m#L339 after adding mixpanel using SPM

It is happening because MIXPANEL_UNIQUE_DISTINCT_ID preprocessor is not defined.

zihejia:
hi @tc-sgupta , thanks for bringing it up. We are thinking of making this configurable through API in addition to the preprocessor. So you will be able to set it when initializing the SDK.

tc-sgupta:
@zihejia Thanks, that would be perfect. Is there any ETA on this?

tc-sgupta:
@zihejia Any update on this issue?

jaredmixpanel:
@tc-sgupta this is available in v4.1.0, we've added a `useUniqueDistinctId` parameter. See: https://github.com/mixpanel/mixpanel-iphone/pull/972

PR #972, opened 2022-01-07
add useUniqueDistinctId

jaredmixpanel:
Since SPM doesn't provide a way to specify pre-processor macros we need to make `MIXPANEL_UNIQUE_DISTINCT_ID` configurable via an API. This PR adds that functionality.

Addresses: https://github.com/mixpanel/mixpanel-iphone/issues/970

### DIFF
--- a/Sources/Mixpanel.m
+++ b/Sources/Mixpanel.m
@@ -1009,7 +1009,7 @@ typedef NSDictionary*(^PropertyUpdate)(NSDictionary*);
 
 - (NSString *)uniqueIdentifierForDevice
 {
-    NSString *distinctId;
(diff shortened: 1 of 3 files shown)


Sources/Mixpanel.m
```diff
@@ -998,6 +1006,20 @@ - (NSString *)macOSIdentifier
 }
 #endif
 
+
+- (NSString *)uniqueIdentifierForDevice
+{
+    NSString *distinctId;
```
zihejia, line 1012:
```suggestion
    NSString *distinctId = nil;
```

zihejia (approved):

PR #977, opened 2022-01-21
Set content-type to application/json as required by Mixpanel ingestion API

zihejia:
set content-type to `application/json` as required by Mixpanel ingestion API


jaredmixpanel (approved):
lgtm

PR #987, opened 2022-04-26
only add $distinct_id to people records

jaredmixpanel:
previously we were putting $distinct_id on all records, $distinct_id is disallowed on groups records, so this was causing them to be rejected by the api


zihejia (approved):
LGTM

Issue #1003, opened 2022-11-04
Only tvOS Pre-Compiled Framework is Available via Github Release and Carthage


csaper:
Hi there @zihejia, @jaredmixpanel  & team -

It appears that only the pre-compiled MixPanel framework for tvOS is available for the latest Release on Github ([download link](https://github.com/mixpanel/mixpanel-iphone/releases/download/v5.0.0/Mixpanel.framework.zip)).

![Pre-Compiled Framework for tvOS Only](https://user-images.githubusercontent.com/316563/200034321-01c3a90b-6c89-4476-8dcc-82808c0b3b05.png)

<hr/>

When attempting to alternately build via Carthage, even when specifying a non-tvOS platform via command line argument (i.e. `carthage build --no-skip-current --cache-builds --platform iOS`), only the tvOS framework appears to be built and archived.

Any way to get an iOS-only .xcframework of Mixpanel v5.0.0?

Thanks!

jaredmixpanel:
@csaper This is fixed in v5.0.1. The Carthage builds attached to the Releases are fixed (and changed to XCFrameworks). You should be able to build an iOS-only XCFramework with `carthage build --platform ios --use-xcframeworks --no-skip-current`

csaper:
Thanks very much @jaredmixpanel!

Issue #997, opened 2022-08-12
Cannot build using Carthage on Xcode 14 beta 5 - schema confusion


m-keller:
**Integration Method:** Carthage
**Xcode Version:** Xcode 14 beta 5
**Library Version:** 4.2.0
**Platform:** iOS
**Language:** Swift
**Description:** 

Carthage correctly builds the iOS target when using Xcode 13. However, on Xcode 14 beta Carthage somehow only finds the "App Extension" schema as the only appropriate build target and that ends here:

```
A shell task (/usr/bin/xcrun xcodebuild -project /Users/REDACTED/ios/Carthage/Checkouts/mixpanel-iphone/Mixpanel.xcodeproj -scheme [App\ Extension]\ Mixpanel -configuration Release CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES archive -showBuildSettings -skipUnavailableActions) failed with exit code 6:
2022-08-12 10:46:48.429 xcodebuild[93169:3580009] [MT] DVTAssertions: ASSERTION FAILURE in /System/Volumes/Data/SWE/Apps/DT/BuildRoots/BuildRoot8/ActiveBuildRoot/Library/Caches/com.apple.xbs/Sources/IDEFrameworks/IDEFrameworks-21330/IDEFoundation/Execution/RunDestinations/IDERunDestinationCLI.m:566
Details:  Unhandled/unexpected case where no run destinations were produced.
Object:   <IDERunDestinationCLI>
Method:   +resolveRunDestinationsWithWorkspace:scheme:buildAction:schemeCommand:schemeTask:destinationSpecifications:architectures:timeout:runDestinationManager:deviceManager:fallbackPreferredSDK:fallbackPreferredArchitectures:skipUnsupportedDestinations:shouldSkipRunDestinationValidation:didDisambiguate:disambiguatedMatches:disambiguatedMatchesDescription:error:
Thread:   <_NSMainThread: 0x6000033b0400>{number = 1, name = main}
Hints: 

Backtrace:
  0   -[DVTAssertionHandler handleFailureInMethod:object:fileName:lineNumber:assertionSignature:messageFormat:arguments:] (in DVTFoundation)
  1   _DVTAssertionHandler (in DVTFoundation)
  2   _DVTAssertionFailureHandler (in DVTFoundation)
[...]

PR #1004, opened 2022-11-09
build xcframeworks with carthage, import targetconditionals, delete app extension scheme

jaredmixpanel:
This PR switches our Carthage build output to XCFrameworks. Per Carthage documentation...

> We encourage using XCFrameworks as of version 0.37.0 (January 2021), and require XCFrameworks when building on an Apple Silicon Mac.
> 
> Multi-architecture platforms are not supported when building framework bundles in Xcode 12 and above. 
> Prefer [building with XCFrameworks](https://github.com/Carthage/Carthage#building-platform-independent-xcframeworks-xcode-12-and-above).

Fixes for...

https://github.com/mixpanel/mixpanel-iphone/issues/997

https://github.com/mixpanel/mixpanel-iphone/issues/1003


zihejia (approved):
LGTM

PR #1010, opened 2023-02-23
Disable watchOS in CI

jaredmixpanel:
Pod linting fails for watchOS in Xcode 14: https://github.com/CocoaPods/CocoaPods/issues/11558


zihejia (approved):
LGTM

Issue #1018, opened 2024-03-16
Support Privacy Manifest in Mixpanel objective C Library


mahavrani:
Dear Mixpanel Support Team, 

Apple has declared that Privacy Manifest support is required in the spring of 2024.

Despite Mixpanel's SDK not being listed among the third-party SDKs on Apple's official list (https://developer.apple.com/support/third-party-SDK-requirements/), I noticed from forum that they received a warning from Apple indicating non-compliance due to the absence of Privacy Manifest support during their submission.

ITMS-91053: Missing API declaration - Your app’s code in the “YourApp” file references one or more APIs that require reasons, including the following API categories: NSPrivacyAccessedAPICategoryUserDefaults. While no action is required at this time, starting May 1, 2024, when you upload a new app or app update, you must include a NSPrivacyAccessedAPITypes array in your app’s privacy manifest to provide approved reasons for these APIs used by your app’s code. For more details about this policy, including a list of required reason APIs and approved reasons for usage, visit: https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api.

Upon further investigation, I noticed that Mixpanel Swift is provided with the necessary support for this requirement. However, there appears to be a gap in support for Objective-C. Given the urgency of the situation, I kindly request that Mixpanel provides the same support for Objective-C at the earliest convenience.

https://github.com/mixpanel/mixpanel-swift/issues/620 - Mixpanel Swift Support link

jaredmixpanel:
`v5.0.5` has been released with the privacy manifest added.

PR #1019, opened 2024-03-18
Add privacy manifest

jaredmixpanel:
Add privacy manifest declaring use of User Defaults

Addresses: https://github.com/mixpanel/mixpanel-iphone/issues/1018


zihejia (approved):
LGTM

PR #1021, opened 2024-03-18
Bump minimum iOS deployment target to 13.0

jaredmixpanel:
Due to: https://stackoverflow.com/questions/77139617/clang-error-sdk-does-not-contain-libarclite-at-the-path


zihejia (approved):
LGTM

PR #1025, opened 2024-03-27
Fix 'Multiple commands produce' error when the App need to add PrivacyInfo.xcprivacy.

ondev:


jaredmixpanel (approved):
Thanks!

jaredmixpanel:
`v5.0.7` is released with this change

